PR #5, opened 2023-05-31
add basic docker compose

ciaranschutte:


compose/.env.schema
```diff
@@ -0,0 +1 @@
+POSTGRES_IMAGE=postgres:9.6
```
joneubank, line 1: Not sure we'll want the postgres version to change between envs... we can probably just hardcode this particular property into the docker compose to keep everyone consistent.

joneubank (approved):

compose/.env.schema
ciaranschutte: sounds good - i agree - just an easy var to get this basic skeleton started

PR #16, opened 2023-06-05
Feat/#965 homepage

ciaranschutte:
- homepage with ActionBox and Overture banner (main ticket work)

additional things to get to endpoint of ticket...
- cleanup from Nextjs boilerplate, removed a lot of unused assets
- adds basic layout with global font loading in
- uikit integration working, including theme extension with types and css prop using next compiler
- add a basic CSS reset
- add a single point of usage for Emotion imports, this may seem redundant but it makes it more straightforward to  change our CSS-in-JS functions if needs be. eg. we can monkey patch a new lib, or augment current functions

localhost_3000_ (1)
localhost_3000_ (2)
localhost_3000_


ciaranschutte:
prettier formatting looking incorrect.
I am investigating.
Should not be a blocker for main code review.

ciaranschutte:
also waiting on links for the action box.

ciaranschutte:
Also also - 
Holding off on uikit release until end of sprint as there are many changes on the fly and will likely also be needed with the upcoming tickets. local install is working for both myself and Dan so this is not a blocker and doesn't distract from the ticketed work to do a release every day.

tsconfig.json
```diff
@@ -1,28 +1,36 @@
...
+		"incremental": true,
+		"jsx": "preserve",
+		"jsxImportSource": "@emotion/react",
+		"plugins": [
+			{
+				"name": "next"
+			}
+		],
+		"paths": {
+			"@/*": ["./src/*"]
```
justincorrigible, line 24: gotta be honest, I'm a bit surprised these are working on the transpiled code. generally I have had to patch typescript in order for the paths to work when not in "dev server" mode. NextJS has really improved quality of life, I guess

src/app/layout.tsx
```diff
@@ -17,20 +17,22 @@
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-import './globals.css';
-import { Inter } from 'next/font/google';
+/**
+ * React.Context, used by ThemeProvider, doesn't work server side so we're defaulting to client side rendering
+ */
+'use client';
```
justincorrigible, line 23: 🤔interesting! this hasn't been necessary in daco nor platform, afaik.
is this a new requirement because of latest Emotion or NextJS?

ciaranschutte: latest NextJS requirement.
Next 13 defaults to everything being treated as a server side component.

tsconfig.json
ciaranschutte: I have not - I will test this case - good call!

Nextjs is honestly doing some insanely awesome work with the dev experience

src/app/layout.tsx
ciaranschutte: React.context just doesn't work in the SSR way of doing things. You're meant to use it for SSR.
It's not a bug or a workaround or anything like that.

ciaranschutte:
converting back to draft PR until dependencies are happy

PR #28, opened 2023-06-16
Feat/#11 footer

ciaranschutte:
Adds Footer component

Related UIKit PR: https://github.com/icgc-argo/uikit/pull/135/files


src/app/components/Footer.tsx
```diff
@@ -0,0 +1,66 @@
...
+				},
+				{
+					displayName: 'Documentation',
+					href: urls.DOCS_URL_ROOT,
+					target: '_blank',
+				},
+				{
+					displayName: 'The Team',
+					href: '',
+				},
```
demariadaniel, line 47: No href or target?

ciaranschutte: doesn't exist yet

demariadaniel (approved):

src/app/components/Footer.tsx
ciaranschutte: in the specification "leave it blank for now" sense of the word - more so than not implemented yet which we could pop a placeholder :)

PR #29, opened 2023-06-18
fix devops

ciaranschutte:


ciaranschutte:
I'm guessing I don't need a release version for this seen as it's just devops? Or should I still increment patch?

justincorrigible:
this pesky little bit will make the build fail if you don't bump the version
https://github.com/icgc-argo/rdpc-ui/blob/develop/Jenkinsfile.groovy#L113

...or alternatively, you could delete this tag here, so it can be created again https://github.com/icgc-argo/rdpc-ui/releases/tag/0.2.0

your call. 👍

ciaranschutte:
@justincorrigible deleted the tag. seems unfair to have a tagged release of something that never even made it to QA :p

justincorrigible (approved):

PR #31, opened 2023-06-19
basic region config

ciaranschutte:


ciaranschutte:
unclear about the best way to structure this - would love some feedback:
```
export const regions = {
	GERMANY: 'Germany',
	CANADA: 'Canada',
} as const;

export type RegionKey = keyof typeof regions; // this creates a dynamic union types GERMANY | CANADA

export const isRegionKey = (value: string): value is RegionKey =>
	regionKeys.some((region) => region === value);
	
export const regionKeys = Object.keys(regions); // this doesn't get automatic typing
```
Two questions: 
1) If we base types off the `regions` object, the regions object itself isn't typed. Is this the way or is it better to explicitly declare a union type of GERMANY | CANADA etc that we add to. I don't imagine it will change much.

2) creating a typeguard of `isRegionKey` seems overkill but it's the only way I could get type checking to work in the config file. an array find didn't have strict enough types.

src/global/config.ts
```diff
@@ -17,10 +17,25 @@
...
+	const envRegionKey = process.env.NEXT_PUBLIC_REGION_KEY || '';
+	const normalizedKey = envRegionKey.toUpperCase().trim();
+
+	// is valid region?
+	if (isRegionKey(normalizedKey)) {
+		return normalizedKey;
+	} else {
+		throw Error(`Invalid region. Must be one of ${regionKeys.join(', ')}`);
+	}
+};
```
joneubank, line 34: I am curious how `NEXT_PUBLIC_REGION_KEY` will be used throughout the app. I suspect this will just be a display string, and if so we don't need to validate its content, just that it is a non-empty string. Therefore also no enum needed.

ciaranschutte: It will be a const of some kind. real example from mocks to display a flag based on region, or a logo based on region `object_of_rdpc_flags[region]`

ciaranschutte: so there is a need for a constant of some kind and a display string. 
I don't like depending on a string to be a constant because you end up string manipulating back and forth for casing

joneubank: Ideally we won't have to make code changes that are specific to any region, this approach bakes in knowledge of the regions and then neccesitates when we add new regions, change a display property of a region, etc.

We should strive as long as possible to keep the code region agnostic. If this can be handled completely with config that would be much easier. If we need to push back on some design ideas that will make all our lives easier long term.

ciaranschutte:
closing as we don't need this level of configuration

PR #35, opened 2023-07-05
Feat 13 contact page

ciaranschutte:
- contact page 
- form using zod schema + react hook form
- recaptcha env var


package.json
```diff
@@ -19,19 +20,24 @@
...
     "react-query": "^3.39.3",
-    "url-join": "^5.0.0"
+    "url-join": "^5.0.0",
+    "zod": "^3.21.4"
   },
-  "peerDependencies": {    
+  "peerDependencies": {
+    "@emotion/react": "^11.11.0",
     "@emotion/styled": "^11.11.0"
   },
```
justincorrigible, line 33: just curious, since this app is not being imported anywhere else, why these need to be `peerDependencies`?
but maybe I'm missing something about this property 🤔

ciaranschutte: theming from uikit works correctly because of however those caches (that you have a far better fix for in arranger I think?) work in the background (pointing to same instances I believe)

src/app/contact/components/info.tsx
```diff
@@ -0,0 +1,170 @@
...
+						variant="sectionHeader"
+						color="secondary"
+						bold
+						as="div"
+					>
+						Downloading Data
+					</Typography>
+					<ul>
+						<li>
+							<Link target="_blank" href={DOCS_DATA_DOWNLOAD_PAGE}>
```
justincorrigible, line 106: fun fact: https://hackernoon.com/unsafe-use-of-target_blank-39413ycf

The ref property should be added to all "_blank"-targetted links
```suggestion
							<Link href={DOCS_DATA_DOWNLOAD_PAGE} rel=”noreferrer noopener” target="_blank" >
```

src/app/contact/components/info.tsx
```diff
@@ -0,0 +1,170 @@
...
+							<Link target="_blank" href={DOCS_SUBMITTING_MOLECULAR_DATA_PAGE}>
+								submit molecular data
+							</Link>
+						</li>
+					</ul>
+				</div>
+
+				<div>
+					<Button variant="secondary">
+						<Link underline={false} target="_blank" href={DOCS_URL_ROOT}>
```
justincorrigible, line 160: link inside a button? is this a nextjs thing?

justincorrigible (approved):

src/app/contact/components/form.tsx
```diff
@@ -0,0 +1,333 @@
...
+					</Row>
+					<Row align="end">
+						<Col>
+							<FormControl required={true} error={!!errors.reCaptcha}>
+								<Controller
+									name="reCaptcha"
+									control={control}
+									rules={{ required: true }}
+									render={({ field }) => {
+										/** @ts-expect-error 3rd party recaptcha comp onChange event type is more closed than react-hook-form onChange */
```
justincorrigible, line 309: 👍

package.json
justincorrigible: 😆 UiKit is just the gift that keeps on giving, uh?

src/app/contact/components/info.tsx
ciaranschutte: No it's a styling uikit thing
not great markup wise (button should be on click action vs a link which is navigational) but it's how we do it design wise

PR #39, opened 2023-07-07
Feat/8 Logout

demariadaniel:
- Adds Dropdown Menu Logout button + removes old button from page
- Moves Logout into Auth Context

https://app.zenhub.com/workspaces/icgc-argo-platform-dk-production-board-5e542d38415f5034e9fed89d/issues/gh/icgc-argo/rdpc-ui/8


src/global/utils/auth.tsx
```diff
@@ -54,18 +55,24 @@ export const storeToken = (egoToken: string) => {
...
-};
-
 export function AuthProvider({ children }: { children: ReactNode }) {
 	const storedToken = getStoredToken();
 	const [egoJwt, setEgoJwt] = useState(storedToken || '');
+	const router = useRouter();
 	const path = usePathname();
-	const initLoginState = path === '/logging-in' && !egoJwt.length ? true : false;
-	const [loggingIn, setLoggingIn] = useState(initLoginState);
+	const loginStateOnPageLoad = path === '/logging-in' && !egoJwt.length;
```
ciaranschutte, line 63: https://github.com/icgc-argo/rdpc-ui/pull/39/files#diff-09e9a1c3d8949b041f6220ac58c961358436d20050a933e1f9bf52a9cd93428cR38

is this line doing the same logic? Is it covered by this line in AuthProvider which covers if a user directly goes to `logging-in` route for some reason

src/app/components/Header.tsx
```diff
@@ -18,15 +18,54 @@
...
+			onClick: async () => {
+				setDropdownOpen(false);
+				logOut();
+			},
+			name: 'Logout',
+			active: false,
+			href: '',
+			LinkComp: Link,
+		},
+	];
```
ciaranschutte, line 66: flawed api to `NavBarElement`, not much improvement without editing uikit
one important thing to note is that `Link` for the `LinkComp` will not work like Next.js native component (no prefetching, SPA routing might be impacted)

having both `href` and `isLink` is redundant - observation on uikit component

src/app/components/Header.tsx
```diff
@@ -43,21 +82,49 @@ const Header = () => {
...
+											/>
+										))}
+									</DropdownMenu>
+								)}
+								<UserBadge
+									showGreeting={true}
+									firstName={'Test'}
+									lastName={'User'}
+									title={'DCC Member'}
+									className={onProfilePage ? 'active' : ''}
```
ciaranschutte, line 113: this is putting the `active` classname on a div
I don't see related styles that this would apply
```javascript

export const UserBadgeContainer = styled('div')`
  text-align: left;
  display: flex;
  align-items: center;
  justify-content: space-between;
  min-width: 200px;
  margin: 0 -24px;
  padding: 0 16px;
  box-sizing: border-box;
`;
```

src/app/components/Header.tsx
```diff
@@ -18,15 +18,54 @@
...
+				logOut();
+			},
+			name: 'Logout',
+			active: false,
+			href: '',
+			LinkComp: Link,
+		},
+	];
+
+	const activeBorder = `solid 3px ${theme.colors.accent1}`;
```
ciaranschutte, line 68: can likely pop this one inline similar to the rest of the styling to keep it consistent

ciaranschutte (approved):
couple of suggestions and a couple of questions
working ok - approved

src/global/constants.ts
```diff
@@ -18,3 +18,4 @@
  */
 
 export const EGO_JWT_KEY = 'EGO_JWT';
+export const LOGIN_NONCE = 'LOGIN_NONCE';
```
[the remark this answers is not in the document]
demariadaniel: This is part of the OIDC Auth Flow to prevent CSRF attacks https://developers.google.com/identity/openid-connect/openid-connect#sendauthrequest
The Cookie is auto-populated by the login flow so I've added logic to remove it on logout

src/app/logging-in/page.tsx
```diff
@@ -30,12 +30,12 @@ export default async function LoggingIn() {
 	const { EGO_API_ROOT, EGO_CLIENT_ID } = getAppConfig();
 	const router = useRouter();
 	const theme = useTheme();
-	const { egoJwt, setEgoJwt, loggingIn, setLoggingIn } = useAuthContext();
+	const { egoJwt, setEgoJwt, authLoading, setAuthLoading } = useAuthContext();
```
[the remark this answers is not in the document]
demariadaniel: Yep!

src/global/utils/auth.tsx
demariadaniel: Ah so I added this to `/logging-in` to prevent a User navigating to /logging-in and then clicking 'Log In' while the request completes. I couldn't get Next server/client to play nicely together here and Header frequently would load Login button on page load, then update to render loading animation. 

I will take another look to see if this can be streamlined but it's definitely something I had to go back-and-forth over several times.

src/app/components/Header.tsx
demariadaniel: Updated

src/app/components/Header.tsx
```diff
@@ -18,15 +18,54 @@
...
+	const profileNavDetails: Array<NavElement> = [
+		{
+			active: profileActive,
+			href: '/landing-page',
+			name: 'Profile & Token',
+			LinkComp: Link,
+		},
+		{
+			isLink: false,
+			onClick: async () => {
```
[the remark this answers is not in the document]
demariadaniel: Removed but need to test when collab is back

src/app/components/Header.tsx
```diff
@@ -43,21 +82,49 @@ const Header = () => {
...
+							</FocusWrapper>
+						) : authLoading ? (
+							<DnaLoader />
+						) : (
+							<LoginButton />
+						)}
+					</AppBarMenuItem>
 				</div>
 			</AppBar>
 		</header>
```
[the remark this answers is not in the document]
demariadaniel: Yes exported as `ProfileMenu`, much cleaner

src/app/components/Header.tsx
demariadaniel: Yes and there's a Type mismatch between modern `next/link` and uikit `Link` so we have to use uikit Link here

ciaranschutte: ah I see! always the freedom to extend uikit - i'll leave that to your discretion :)

src/app/components/Header.tsx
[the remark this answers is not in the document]
demariadaniel: I've moved this into a separate `ProfileMenu` component

src/app/logging-in/page.tsx
```diff
@@ -49,7 +49,7 @@ export default async function LoggingIn() {
 				const newToken = await res.text();
 				storeToken(newToken);
 				setEgoJwt(newToken);
-				setLoggingIn(false);
+				setAuthLoading(false);
 			})
 			.catch(console.error);
 	});
```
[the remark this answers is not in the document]
demariadaniel: Yes - beyond scope of this ticket which is why I left it as just `console.error`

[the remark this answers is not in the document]
demariadaniel: Multiple Platform utils using a `catch -> console -> custom error handler` pattern

src/app/components/Header.tsx
demariadaniel: Removed, looks fine

PR #41, opened 2023-07-10
Feat/contact style

ciaranschutte:


src/app/components/ThemeProvider.tsx
```diff
@@ -50,7 +50,6 @@ export default function ThemeProvider({ children }: { children: React.ReactNode
 
 					body {
 						line-height: 1.5;
-						-webkit-font-smoothing: antialiased;
```
ciaranschutte, line 53: too experimental - not recommended

src/app/page.tsx
```diff
@@ -141,11 +141,10 @@ const ActionBox: ComponentType = () => (
 
 export default function Home() {
 	return (
-		<main>
+		<>
```
justincorrigible, line 144: this is bound to get messy for other pages sharing the same layout, with nested mains, etc. no?

ciaranschutte: Yes I think so that's why I moved the main element to the root layout
Not header, not footer, it's main - we could narrow what "main" means in our dashboard context more but "not header, not footer" is a good start I think

layouts nest

justincorrigible: my concern here is others not using a fragment as the root of the page, then it's a "double main" etc.

ciaranschutte: ah i see i see - updated

agree. worse to have multiple mains, or terrible markup nesting than to not have one at all.

justincorrigible: yeah. I'm not loving this new layout vs page thing in nextjs

PR #42, opened 2023-07-10
clean up footer

ciaranschutte:


justincorrigible (approved):

src/app/components/Footer.tsx
```diff
@@ -19,49 +19,130 @@
...
+import rdpcLogo from '/public/assets/rdpc-logo.svg';
 
 const { UI_VERSION, REGION } = getAppConfig();
 const subtitle = `RDPC ${REGION} Clinical Data Submission Portal - ${UI_VERSION}`;
 
-const Logo = () => <div>RDPC</div>;
+const Logo = () => <Image src={rdpcLogo} alt="RDPC logo" />;
 
-export default function GlobalFooter() {
+const links = [
```
justincorrigible, line 33: this could go into a separate (local) constants file

src/app/components/Footer.tsx
```diff
@@ -19,49 +19,130 @@
...
+						padding-left: 22px;
+					`}
+				>
+					{/** nav links */}
+					<div
+						css={css`
+							margin-right: 22px;
+						`}
+					>
+						{links.map(({ displayName, href, target }, idx) => (
```
justincorrigible, line 122: snark for comedic tone, not for snub:
how is idx better than index? "3 > 5" economy?

src/app/components/Footer.tsx
```diff
@@ -19,49 +19,130 @@
...
+									target={target}
+									href={href}
+									rel={target === '_blank' ? 'noreferrer noopener' : ''}
+									css={css`
+										color: ${theme.colors.primary};
+									`}
+								>
+									{displayName}
+								</Link>
+								{idx !== links.length - 1 && (
```
justincorrigible, line 134: having a slash divider be an icon and adding it through array index logic seems like an overengineered alternative to an easy CSS solution

ciaranschutte: I agree

PR #47, opened 2023-07-18
Feat/#14 programs list page

ciaranschutte:
NB: part of larger ticket 

- adds `Submission` page
- adds skeleton layout components
- adds `ProgramList` table

Notes:
- types are open to be narrowed when data hookup ticket is complete


src/app/submission/components/ProgramList.tsx
```diff
@@ -0,0 +1,60 @@
...
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+'use client';
+
+import { useTheme } from '@/lib/emotion';
+import { Table, Typography, css } from '@icgc-argo/uikit';
+import { columns } from '../tableConfig';
+
+export default function ProgramList({ programs }: { programs: any[] }) {
```
ciaranschutte, line 25: will define narrow type once data hookup ticket is done

src/app/submission/components/Sidemenu.tsx
```diff
@@ -0,0 +1,23 @@
...
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+'use client';
+
+export default function SideMenu({ programs }: { programs: any[] }) {
```
ciaranschutte, line 21: same here. will narrow type when data hookup ticket is done

src/app/submission/tableConfig.tsx
```diff
@@ -0,0 +1,138 @@
...
+	},
+	{
+		header: 'Cancer Types',
+		accessorKey: 'cancerTypes',
+		cell: ({ row: { original } }) => (
+			<div>
+				{original.cancerTypes.map((cancerType, i) => (
+					<div key={cancerType}>
+						{cancerType}
+						{i < original.cancerTypes.length - 1 && ','}
```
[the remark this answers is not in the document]
ciaranschutte: like `cancerTypes.join(',')` ? 
the use case for the map is to return the div which is a block element so in the ui you get stacked list

.gitignore
```diff
@@ -34,3 +34,6 @@ yarn-error.log*
 # typescript
 *.tsbuildinfo
 next-env.d.ts
+
+# generic name for files that shouldn't be checked in (handy for mock data)
+*.temp*
```
justincorrigible, line 39: Nice one, mate! 👍

src/app/layout.tsx
```diff
@@ -22,9 +22,9 @@
  */
 'use client';
 
+import { AuthProvider } from '@/global/utils/auth';
 import { ReactNode } from 'react';
 import { QueryClient, QueryClientProvider } from 'react-query';
-import { AuthProvider } from '@/global/utils/auth';
```
justincorrigible, line 27: nitpick: what's this change's purpose. was it done by https://www.npmjs.com/package/prettier-plugin-organize-imports

src/app/submission/tableConfig.tsx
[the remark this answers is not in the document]
justincorrigible: looks like a "display only" sort of deal, so maybe it'd be better to use css 
```
 :not(:last-of-type)::after {
   content: ",";
}
```

src/app/layout.tsx
ciaranschutte: either that or a merge conflict with the cleanup

justincorrigible: well, I may suggest adding that package to all projects, for good measure

ciaranschutte: i'm liking the package a lot - very useful

justincorrigible: works fairly well. wish it also added empty lines between import groups, but this does the job well enough

ciaranschutte: I dislike the empty lines because it's so much a developer preference... loads of places where there's just empty lines for no reason

justincorrigible: To give a brief example of what I mean to me whitespace emphasises separation of ideas the same way in English we use puntuation in sentences to prevent the "run-on" ones as well as how we use empty lines between separate paragraphs to separate topics etc because if we wrote code the same way I'm writing this it'd still be understandable but god does it take a lot more effort don't you agree?


😂

ciaranschutte: 100% agree. unfortunately like most of these topics unless there's a standard way to enforce it , it soon goes to chaos

justincorrigible: you're right, that's the rub.
I'll keep testing other prettier plugins but the imports order being respected is certainly an improvement. 💯

justincorrigible (approved):
Hoping to see that prettier imports plugin in the next PR

ciaranschutte:
> Hoping to see that prettier imports plugin in the next PR

https://github.com/icgc-argo/rdpc-ui/blob/develop/package.json#L49 it's in dev :)

PR #59, opened 2023-08-02
test

ciaranschutte:
draft pr - can add docs, clean up , add copyright, fix spellings etc if this works ok for team

big picture: call api route to get access to runtime vars

tried several approaches:
- regular js module I couldn't get to work with app due to fetch being async
- using a hook works and comes close but I want one instance of the data 
- ended up using context, shared instance of data (as close to a single module export of data I could get)


src/app/api/route.ts
```diff
@@ -0,0 +1,26 @@
...
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { NextResponse } from 'next/server';
+import { headers } from 'next/headers';
+
+export async function GET() {
+	headers()
+	return NextResponse.json({ REGION: process.env.CIARAN_RUNTIME_REGION });
```
justincorrigible, line 25: 🤔 !!!

ciaranschutte: You're going to have to give me more than that I'm afraid.

src/global/useConfig.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export const ConfigProvider = ({ children }: { children: any }) => {
+	// only need to do this once to get values from runtime api route
+	const [config, setConfig] = useState({});
+	useQuery('config', async () => {
+		const res = await fetch('http://localhost:3000/api');
+		const data = await res.json();
+		console.log('use query', data);
+		setConfig(data);
+	});
```
demariadaniel, line 14: Will this need a dependency array?

src/global/useConfig.tsx
```diff
@@ -0,0 +1,18 @@
+import { createContext, useContext, useState } from 'react';
+import { useQuery } from 'react-query';
+
+const ConfigContext = createContext({});
+
+export const ConfigProvider = ({ children }: { children: any }) => {
+	// only need to do this once to get values from runtime api route
+	const [config, setConfig] = useState({});
+	useQuery('config', async () => {
+		const res = await fetch('http://localhost:3000/api');
```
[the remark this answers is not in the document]
ciaranschutte: for sure - good call! I'm fairly certain the relative '/api' actually works seen as it's the same server instance client pages are rendered from.

src/global/useConfig.tsx
```diff
@@ -0,0 +1,18 @@
+import { createContext, useContext, useState } from 'react';
+import { useQuery } from 'react-query';
+
+const ConfigContext = createContext({});
+
+export const ConfigProvider = ({ children }: { children: any }) => {
```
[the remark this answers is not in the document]
ciaranschutte: No, context providers in general have difficulty with SSR. 
So this is the purely client approach to getting server side values. 
To get any values we'd add them to the api route (which is server side)

That's where all this difficulty lies. client vs server aspects.

src/global/useConfig.tsx
ciaranschutte: This is using useQuery from react-query so it works a little different to useEffect
it also caches by default

that been said we do have to tweak some work in `layout.tsx` having multiple re-renders that aren't needed.
This is an existing issue already.

ciaranschutte: I hear what you're saying - ensure as much as possible it gets called once. once is all that's needed

src/app/api/route.ts
justincorrigible: oh I've meant that as a "wow, let me think about this one very clever solution!". sorry, lol

src/global/useConfig.tsx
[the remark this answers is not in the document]
justincorrigible: hm so in a way, at risk of raining on the parade, this doesn't solve the problem for server routines that rely on env values, but I guess for those we can just call the "api" route directly... 🤔 messy, but could work

src/global/useConfig.tsx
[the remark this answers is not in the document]
justincorrigible: I am not sure the server-side will be able to tell what the browser route/path will be, so that may not work.

however, surely we can just pass that as an env value, once we figure out that part of this thing, right?

oh wait, chicken and egg!! lol

src/app/api/route.ts
ciaranschutte: all good :D

src/global/useConfig.tsx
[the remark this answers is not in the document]
ciaranschutte: not raining on the parade at all. this is exactly why I'm talking so much about this.

not 100% sure what "server routines" in this context is but in general it appears that:
server has access runtime process env
client side only has access to env vars from build time, because they all get inlined, including process.env

PR #61, opened 2023-08-07
Runtime client vars

ciaranschutte:
- runtime config  
  - add api route
  - add central config point that doesn't completely break convention
  - add context provider
  - update all usages of env vars
- add env var to denote when NEXT is building
- update readme
- refactor global url. turns out there's very little "global" about them. better to move them to where they're used instead of changing this into a hook.


src/app/api/config/config.ts
```diff
@@ -0,0 +1,64 @@
...
+			serverEnv.NEXT_PUBLIC_EGO_PUBLIC_KEY || process.env.NEXT_PUBLIC_EGO_PUBLIC_KEY || '',
+		UI_VERSION: packageJSON.version,
+		REGION: serverEnv.NEXT_PUBLIC_REGION || process.env.NEXT_PUBLIC_REGION || '',
+		PLATFORM_UI_ROOT:
+			serverEnv.NEXT_PUBLIC_PLATFORM_UI_ROOT || process.env.NEXT_PUBLIC_PLATFORM_UI_ROOT || '',
+		RECAPTCHA_SITE_KEY:
+			serverEnv.RECAPTCHA_SITE_KEY || process.env.NEXT_PUBLIC_RECAPTCHA_SITE_KEY || '',
+	};
+	return config;
+};
```
ciaranschutte, line 64: could use some cleanup but it's functional.
dynamic keys become an issue because of build time inlining of variables.

src/lib/egoJwt.ts
```diff
@@ -21,9 +21,8 @@
 import memoize from 'lodash/memoize';
 
 import createEgoUtils from '@icgc-argo/ego-token-utils';
-import { getAppConfig } from '@/global/config';
 
-const TokenUtils = createEgoUtils(getAppConfig().EGO_PUBLIC_KEY);
+const TokenUtils = createEgoUtils('');
 
 export const decodeToken = memoize((egoJwt?: string) =>
 	egoJwt ? TokenUtils.decodeToken(egoJwt) : null,
```
ciaranschutte, line 28: this a little more involved to get working with the new app config.
none of these functions appear to be in use, so may be a non issue
@demariadaniel are these needed?

ciaranschutte:
investigating CI issues

src/lib/egoJwt.ts
ciaranschutte: changed to empty string to not block, did not remove until verifying not needed

src/app/api/config/config.ts
```diff
@@ -0,0 +1,79 @@
...
+		UI_VERSION: packageJSON.version,
+		REGION: serverEnv.NEXT_PUBLIC_REGION || process.env.NEXT_PUBLIC_REGION || '',
+		PLATFORM_UI_ROOT:
+			serverEnv.NEXT_PUBLIC_PLATFORM_UI_ROOT || process.env.NEXT_PUBLIC_PLATFORM_UI_ROOT || '',
+		RECAPTCHA_SITE_KEY:
+			serverEnv.RECAPTCHA_SITE_KEY || process.env.NEXT_PUBLIC_RECAPTCHA_SITE_KEY || '',
+		ARGO_ROOT: serverEnv.ARGO_ROOT || process.env.ARGO_ROOT || 'https://www.icgc-argo.org',
+		EGO_API_ROOT,
+		EGO_CLIENT_ID,
+		EGO_LOGIN_URL,
```
demariadaniel, line 74: Ego vars should be treated the same no?

ciaranschutte: in regards to the `serverr || client || default`? yes - they're defined just above this block so we can cleanly put in the `EGO_LOGIN_URL` which is composed from other variables. only other way to do that is a function expression object I think and that looks pretty gnarly suddenly throwing `this` in here

Dockerfile
```diff
@@ -12,7 +12,7 @@ WORKDIR /appDir
 
 COPY . .
 RUN npm ci
-RUN npx next build
+RUN npm run build
```
demariadaniel, line 15: Why the change from npx next to npm run?

src/lib/egoJwt.ts
demariadaniel: I see what you mean, this was setting up for additional Login functionality but aren't in use at the moment

demariadaniel (approved):

PR #64, opened 2023-08-10
rc/0.5.0

ciaranschutte:


package.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "rdpc-ui",
-  "version": "0.4.0",
+  "version": "0.5.0",
   "private": true,
   "scripts": {
     "dev": "next dev",
-    "build": "next build",
+    "build": "NEXT_IS_BUILDING=true next build",
```
justincorrigible, line 7: ooh new flag. they just love polluting the env, eh?

ciaranschutte: that's me polluting the env ;)

ciaranschutte: and yes I do enjoy it :p

PR #74, opened 2023-08-20
Data setup

ciaranschutte:
- large amount of files are from updating the import statement for App config provider
- adds GQL Apollo handling
- adds typechecking with generated types
- adds auth for handling gql requests
- adds basic loading and error handling - to be improved https://github.com/icgc-argo/rdpc-ui/issues/77
- hydration issues are an existing bug https://github.com/icgc-argo/rdpc-ui/issues/78


codegen.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+import { CodegenConfig } from '@graphql-codegen/cli';
+import dotenv from 'dotenv';
+
+// point to .env.local because .env in nextjs is a default committed file
+dotenv.config({ path: '.env.local' });
+
+const gqlConfig: CodegenConfig = {
+	schema: `${process.env.NEXT_PUBLIC_GATEWAY_API_ROOT}/graphql`,
```
ciaranschutte, line 27: this file is outside the context of Next but it makes sense to try and at least keep consistent naming/file instead of having multiple env configurations depending on process context

src/app/(post-login)/submission/types.ts
```diff
@@ -0,0 +1 @@
+export type Programs = any;
```
demariadaniel, line 1: Can we update this?

demariadaniel (approved):
Huge! Looking great

PR #88, opened 2023-09-21
Submit clinical file records

ciaranschutte:
Most of clinical submission page
Only new code is https://github.com/icgc-argo/rdpc-ui/pull/88/files#diff-7382a2577603d2cc3116410d5faec9bb2dbae4bc6ad09dc99b75067cebaf8ba6R65 which is for additional debugging

removed commented out code and console logs for PR

merging into feature branch


ciaranschutte:
merging into feature branch (not needing the same level of approval as dev/main in my opinion)

PR #97, opened 2023-10-07
permissions: user role check

ciaranschutte:
- adds a user role check permission hook
- calls it from AuthHook but kept as a separate hook intentionally

Change from `platform-ui` functionality: 
utility functions for individual roles created within the hook right now
this is because we need to use `appConfig` hook to create token utils. before, we created this statically with env vars.
we can change this as an enhancement to either use `useCallback` so they won't get recreated with multiple renders or pass in token utils as a var with DI and move the functions out of our hook.


ciaranschutte:
very out of date.

PR #100, opened 2023-10-26
Feat/38 submission approval screen

demariadaniel:


package.json
```diff
@@ -70,6 +70,12 @@
     "**/*": "prettier --write --ignore-unknown"
   },
   "prettier": {
+    "printWidth": 100,
+    "semi": true,
+    "singleQuote": true,
+    "trailingComma": "all",
+    "useTabs": true,
+    "tabWidth": 2,
```
demariadaniel, line 78: Moved Prettier rules to package.json as prettierrc was getting ignored, causing unnecessary Git changes
LMK if there's an alternate preferred way to do this, but Platform UI is set up w/ rules in package.json

src/app/(post-login)/submission/layout.tsx
```diff
@@ -32,7 +32,7 @@ export default function AppLayout({ children }: { children: ReactNode }) {
 		<div
 			css={css`
 				display: grid;
-				grid-template-columns: ${isSidebarActive ? '248px' : '40px'} 1fr;
+				grid-template-columns: ${isSidebarActive ? '248px 86vw' : '40px 98vw'};
```
demariadaniel, line 35: Insure Submission content fits in view window

ciaranschutte: I haven't seen mixing of px and view port units before - interesting!
keep in mind `vw` and `vm` are percents and `px` are absolute

src/app/(post-login)/submission/program/[shortName]/clinical-submission/components/ProgressBar.tsx
```diff
@@ -98,11 +101,22 @@ const ProgressBar: FC<ProgressBarProps> = ({
       ? "success"
       : "disabled",
   };
+  const pendingApprovalWidth = `${approvalBarWidth || 100}px`;
```
Comment (ciaranschutte) [inline src/app/(post-login)/submission/program/[shortName]/clinical-submission/components/ProgressBar.tsx:104]: Is this something that needs to be a prop to `ProgressBar` comp? 
seems like it may be a case of "pending approval" progress item is always larger

src/global/utils/auth.tsx
```diff
@@ -16,86 +16,105 @@
...
+    setAuthLoading,
+    logIn,
+    logOut,
+    permissions,
+    TokenUtils,
+  };
+
+  return (
+    <AuthContext.Provider value={value}>
+      <Suspense
```
ciaranschutte, line 106: there are `Suspense` boundary errors on the page and this is the only place in the code where we use suspense boundaries. I wonder if there could be a fix here?

src/global/utils/auth.tsx
```diff
@@ -16,86 +16,105 @@
...
 
 const AuthContext = createContext<AuthContextValue>({
-	egoJwt: '',
-	setEgoJwt: () => '',
-	authLoading: false,
-	setAuthLoading: () => false,
-	logIn: () => undefined,
-	logOut: () => undefined,
+  egoJwt: "",
+  setEgoJwt: () => "",
```
ciaranschutte, line 51: I don't think this returns a string as type
you're on the right track with the react dispatch generics

src/global/utils/auth.tsx
```diff
@@ -16,86 +16,105 @@
...
+  authLoading: false,
+  setAuthLoading: () => false,
+  logIn: () => undefined,
+  logOut: () => undefined,
+  permissions: [],
+  TokenUtils: createEgoUtils(""),
```
ciaranschutte, line 57: are these odd placeholder functions needed or can you pass the actual functions?

ciaranschutte (changes requested):

ciaranschutte:
@demariadaniel couple of suggestions. assuming you have the prettier chaos under control from your slack message.

one point. before approval re the context functions. @justincorrigible 2nd opinion please? I may be overlooking something obvious

src/global/utils/auth.tsx
demariadaniel: This is from the very early days of setting up the app + Login and definitely worth a revisit esp b/c of the errors

package.json
justincorrigible: personally I prefer having separate files for separate concerns...

The manifest file (package.json) tends to get quite busy as it is, in most projects.

In case it helps, `prettier.config.js` works fairly well in Arranger. https://github.com/overture-stack/arranger/blob/develop/prettier.config.js

src/app/(post-login)/submission/layout.tsx
justincorrigible: @ciaranschutte the unit mix makes sense if you see height in relation to a "constant" font size (8'd have opted for rem) vs viewport width depending on browser being resized

.prettierrc.json
```diff
@@ -1,8 +1,8 @@
...
-	"semi": true,
-	"singleQuote": true,
-	"trailingComma": "all",
-	"useTabs": true,
-	"tabWidth": 2
+  "printWidth": 100,
+  "semi": true,
+  "singleQuote": true,
+  "trailingComma": "all",
+  "useTabs": true,
```
justincorrigible, line 6: The irony of seeing the prettier configs file being reformatted without considering the rules specified therein (tabs)
😂

src/app/(post-login)/submission/layout.tsx
ciaranschutte: wouldn't this potentially result in overflow/underflow though? 
if 248px is not 14% (100 - 86) of the view port width  .. that's one col of fixed size, and another col of fixed size that is 86% of viewport. 

are you talking about that idea of setting a base font size and setting everything, including layout dimensions, relative to it?

src/app/components/Header.tsx
```diff
@@ -16,113 +16,115 @@
  * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
-'use client';
+"use client";
```
justincorrigible, line 19: oof. prettier is making fun of us in this PR. uh?

src/app/(post-login)/submission/layout.tsx
justincorrigible: oh, I was strictly focused on the unit mix... the actual values, I have no clue about 🤷‍♂️

ciaranschutte: @demariadaniel  not sure where those numbers came from but could we change this to, otherwise the page gets a little overflowing:
```
 grid-template-columns: ${isSidebarActive
          ? "248px calc(100vw - 248px)"
          : "40px calc(100vw - 40px)"};
```

justincorrigible: Oh. I see what you meant!
Yeah, this suggesrion certainly seems more robust. 

I'd take it a step further by abstracting those constants into reusable variables, that also go into the sidebar itself (so if the number changes at some point, it does it in both places at once)

ciaranschutte: <insert "100" emoji>

src/global/utils/auth.tsx
justincorrigible: it seems the code was already there, and the change is just a prettier proble..

while I agree with the suggestion, I think it's way more urgent to amend the PR with Prettier working.

It's difficult to parse (or approve) with all these styling regressions.

ciaranschutte: good point. I didn't check if it was new code or not as my instinct is "see green, comments are fine"

justincorrigible:
I may recommend we get back to reviewing this PR once it's been amended with a working Prettier. if others are ok approving it as is, go ahead, but I don't feel comfortable doing so as is.

.prettierrc.json
justincorrigible: for clarity, comment not intended as anything else but to point out I find this funny.

src/app/(post-login)/submission/layout.tsx
demariadaniel: Great catch @ciaranschutte this did indeed overflow + calc is exactly what I was looking for

For context, I added the `vw` changes to contain the table within the page:

Screenshot 2023-10-27 at 12 11 44 PM
Screenshot 2023-10-27 at 12 11 34 PM

demariadaniel: Using vw the table fits on the page:
Screenshot 2023-10-27 at 12 10 42 PM

package.json
demariadaniel: Fixed in: https://github.com/icgc-argo/rdpc-ui/pull/102

src/app/(post-login)/submission/program/[shortName]/clinical-submission/components/ProgressBar.tsx
```diff
@@ -40,9 +41,14 @@ const checkEntities = (clinicalEntities: ProgressBarProps['clinicalEntities']) =
 type ProgressBarProps = {
 	clinicalEntities: ClinicalSubmission['clinicalEntities'] | undefined;
 	clinicalState: ClinicalSubmissionQuery['clinicalSubmissions']['state'] | undefined;
+	approvalBarWidth?: number;
 };
 
-const ProgressBar: FC<ProgressBarProps> = ({ clinicalEntities, clinicalState }) => {
+const ProgressBar: FC<ProgressBarProps> = ({
```
Comment (justincorrigible) [inline src/app/(post-login)/submission/program/[shortName]/clinical-submission/components/ProgressBar.tsx:47]: Going to take a wager, and assume the use of FC here is so you can pass children?
I recall ages ago using the FC type was discouraged, because specifying the type explicitly leads to less errors.

you could use `PropsWithChildren<Props>` if children is all you need

src/app/(post-login)/submission/program/[shortName]/clinical-submission/components/Header.tsx
```diff
@@ -61,8 +62,14 @@ const Header: FC<HeaderProps> = ({
     DOCS_URL_ROOT,
     "/docs/submission/submitting-clinical-data",
   );
+  const { egoJwt, permissions, TokenUtils } = useAuthContext();
 
-  const isDcc = false;
+  const isDcc = useMemo(() => TokenUtils.isDccMember(permissions), [egoJwt]);
+  const isRdpc = useMemo(
+    () => (egoJwt ? TokenUtils.isRdpcMember(permissions) : false),
+    [egoJwt],
```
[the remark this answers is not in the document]
Comment (demariadaniel) [reply src/app/(post-login)/submission/program/[shortName]/clinical-submission/components/Header.tsx]: I believe it would (and tested it to be sure), updated!

src/global/utils/auth.tsx
demariadaniel: @ciaranschutte Give this a second look after Prettier revisions. These are just initializing the context, actual functions are passed in the body of AuthProvider. I updated syntax to use `() => {}` instead.

src/app/(post-login)/submission/program/[shortName]/clinical-submission/components/ProgressBar.tsx
Comment (demariadaniel) [reply src/app/(post-login)/submission/program/[shortName]/clinical-submission/components/ProgressBar.tsx:47]: FC<ProgressBarProps> is not my doing. In platform-ui this is `ComponentType<ProgressBarProps>`, which works fine in RDPC as well.
`PropsWithChildren<ProgressBarProps>` does not work for me.

src/app/(post-login)/submission/program/[shortName]/clinical-submission/components/ProgressBar.tsx
Comment (demariadaniel) [reply src/app/(post-login)/submission/program/[shortName]/clinical-submission/components/ProgressBar.tsx]: This is a carryover from platform-ui
It is used here: https://github.com/icgc-argo/platform-ui/blob/develop/components/pages/submission-system/program-dashboard/ProgramWorkspaceStatus/index.tsx#L142

src/app/(post-login)/submission/program/[shortName]/clinical-submission/components/ProgressBar.tsx
Comment (justincorrigible) [reply src/app/(post-login)/submission/program/[shortName]/clinical-submission/components/ProgressBar.tsx:47]: you're right. once again I feel into the "it's green, new code" trap 😂

in either case, at risk of stating the obvious and just to verify a suspicion: the replacement for 
`const X: FC<Props> = props =>...` is `const X = (props: PropsWithChildren<Props>) =>...`, which in turn lets TS infer the return type (a much better way to use TS in React)

is that how you did it?

i.e. ' const X: PropsWithChildren<Props> =...` wouldn't work because X is meant to be a component, not just its props.

Comment (demariadaniel) [reply src/app/(post-login)/submission/program/[shortName]/clinical-submission/components/ProgressBar.tsx:47]: OK I see. `const X = (props: PropsWithChildren<Props>) =>` does work.

Comment (justincorrigible) [reply src/app/(post-login)/submission/program/[shortName]/clinical-submission/components/ProgressBar.tsx]: glad to hear! thanks for helping leave these systems better than we found them 🙏

ciaranschutte:
merging based on slack "all PR addressed" comment 👍

PR #102, opened 2023-10-27
🐛 Prettier Config

demariadaniel:
- Moves prettier import out of package.json to prevent overriding prettierrc (https://prettier.io/docs/en/configuration.html)
- Reformats files in repo
- Relates to discussion on https://github.com/icgc-argo/rdpc-ui/pull/100

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -4,5 +4,6 @@
 	"singleQuote": true,
 	"trailingComma": "all",
 	"useTabs": true,
-	"tabWidth": 2
+	"tabWidth": 2,
+	"plugins": ["prettier-plugin-organize-imports"]
 }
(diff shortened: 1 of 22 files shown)


.prettierrc.json
```diff
@@ -4,5 +4,6 @@
 	"singleQuote": true,
 	"trailingComma": "all",
 	"useTabs": true,
-	"tabWidth": 2
+	"tabWidth": 2,
+	"plugins": ["prettier-plugin-organize-imports"]
```
justincorrigible, line 8: This shouldn't be necessary, as the VSCode extension picks plugins up automatically

justincorrigible (approved):
so much prettier! 

ready to squash and merge

.prettierrc.json
demariadaniel: Oh so `plugins` can be removed?

PR #108, opened 2023-11-01
Page permissions

ciaranschutte:


ciaranschutte:
https://github.com/icgc-argo/rdpc-ui/pull/108/files#diff-484f563422813ea468a6ce776b24d347c5dd72d9e792d0e2b1d1786992a7fdd6R456
main part to take a look at. draft stage. will clean up / move to it's own file etc.

src/app/(post-login)/submission/program/[shortName]/clinical-submission/page.tsx
```diff
@@ -451,4 +452,9 @@ const ClinicalSubmissionPage = ({ params: { shortName } }: { params: { shortName
 	}
 };
 
-export default ClinicalSubmissionPage;
+export default pageWithPermissions(ClinicalSubmissionPage as React.ComponentType, [
```
Comment (ciaranschutte) [inline src/app/(post-login)/submission/program/[shortName]/clinical-submission/page.tsx:455]: usage of `as` due to a rabbit hole with EmotionJSX.Element typings
tried to use `React.ComponentType` directly but we are using the Emotion compiler for our tsx. This means all our components return `EmotionJSX.Element | undefined`. I can get the typings working until the `pageWithPermissions` higher order function, but it complains on rendering `<Page>`

src/app/(post-login)/submission/components/ProgramMenu.tsx
```diff
@@ -74,101 +135,51 @@ const parseGQLResp = (data: SideMenuQuery) => {
...
 				<MenuItem
-					level={2}
-					content="All Programs"
-					onClick={setActiveProgram(-1)}
-					selected={pathname === '/submission'}
+					level={3}
+					content={
+						<StatusMenuItem>
+							Register Samples
+							{isSubmissionSystemDisabled ? (
```
demariadaniel, line 162: Just my opinion but I feel like we should move this logic out of render into its own variable given its a nested ternary checking several different conditions

`const StatusIcon = isSubmissionSystemDisabled ? (
		<Icon name="lock" fill="accent3_dark" width="15px" />
	) : programStatusData?.clinicalRegistrationHasError ? (
		<Icon name="exclamation" fill="error" width="15px" />
	) : programStatusData?.clinicalRegistrationInProgress ? (
		<Icon name="ellipses" fill="warning" width="15px" />
	) : null;`
	
	`<StatusMenuItem>
	    Register Samples
	    {StatusIcon}`

src/app/403.tsx
```diff
@@ -0,0 +1,93 @@
...
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+ * OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+ * SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ * INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
```
demariadaniel, line 18: Two issues from testing:

a) attempting to navigate to `register` and `submit` pages is causing 403 errors
b) `/403` page is not found so I'm landing on the 404 page

src/app/(pre-login)/logging-in/page.tsx
```diff
@@ -31,7 +31,7 @@ export default async function LoggingIn() {
 	const { egoJwt, authLoading, setAuthLoading, logIn } = useAuthContext();
 	const EGO_TOKEN_URL = `${EGO_API_ROOT}/api/oauth/ego-token?client_id=${EGO_CLIENT_ID}`;
 
-	if (egoJwt) router.push('/landing-page');
```
demariadaniel, line 34: We can also remove the /landing-page component now that this is working

src/app/(post-login)/submission/components/ProgramMenu.tsx
```diff
@@ -74,101 +135,51 @@ const parseGQLResp = (data: SideMenuQuery) => {
 	};
 };
 
-const ProgramMenu = ({ searchQuery }: { searchQuery: string }) => {
-	const params = useParams();
+const MenuContent = ({ programName }: { programName: string }) => {
 	const pathname = usePathname();
-	const activeProgramName = typeof params.shortName !== 'string' ? '' : params.shortName;
+	const pathnameLastSegment = pathname.split('/')[-1];
```
demariadaniel, line 140: I don't think this is working as intended, [-1] is the same as `pathname.split().-1` and returns `undefined`

src/app/(post-login)/submission/program/[shortName]/clinical-submission/page.tsx
Comment (ciaranschutte) [reply src/app/(post-login)/submission/program/[shortName]/clinical-submission/page.tsx]: https://github.com/icgc-argo/rdpc-ui/pull/108/commits/9e68132df235548f829b056e7b1d482949fc07fb

```typescript
type Component = (...args: any[]) => EmotionJSX.Element; // this works as a component type
```
still some kinks to iron out. Typescript doesn't like implicit returns in functions. This can be toggled as a config but I'd rather try and make it work first. so if we have Components operating with implicit returns like this:
```typescript
type Component = (...args: any[]) => EmotionJSX.Element | undefined; 
```
Typescript won't like the render call when it comes to `<Component/>`
There's probably a way to call a component function to check for `undefined` return type but this seems incorrect to me. 

The way around it is no implicit returns but then you account for all cases and can end up with ugly filler: 
https://github.com/icgc-argo/rdpc-ui/pull/108/commits/9e68132df235548f829b056e7b1d482949fc07fb#diff-484f563422813ea468a6ce776b24d347c5dd72d9e792d0e2b1d1786992a7fdd6R453

src/app/components/Page.tsx
```diff
@@ -0,0 +1,55 @@
...
+type PageProps = {
+	Component: React.ComponentType;
+	acceptedRoles: AcceptedRole[];
+	// Nextjs dynamic routing params
+	urlParams: any;
+};
+
+const Page = ({ Component, acceptedRoles, urlParams }: PageProps) => {
+	const { egoJwt } = useAuthContext();
+	const programName = urlParams.params.shortName;
```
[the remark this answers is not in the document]
ciaranschutte: always good to post examples Jon thanks for asking, always happy to have more info :)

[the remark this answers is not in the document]
ciaranschutte: https://github.com/icgc-argo/rdpc-ui/pull/108/commits/11574e951d7985afe25c97da905b381fe70985fc

How does this look? 
accounts for all variations including `undefined` https://nextjs.org/docs/app/api-reference/file-conventions/page#searchparams-optional

downside is there's some logic order I dislike.... but that's hooks for you. need all the conditionals after the hooks.

src/app/(post-login)/submission/components/ProgramMenu.tsx
ciaranschutte: You are correct - not working as intended.
Had it my head I could negative value it like `slice`
Only positive numbers work in array lookup like this

src/app/(pre-login)/logging-in/page.tsx
ciaranschutte: https://github.com/icgc-argo/rdpc-ui/pull/108/commits/286e6e8789cff1ac5fea29b0b262aaf2653f4042

src/app/(post-login)/submission/components/ProgramMenu.tsx
ciaranschutte: https://github.com/icgc-argo/rdpc-ui/pull/108/commits/26a42b9b392a7eb1625cbdc13f1c86e9a5ef87b4

src/app/(post-login)/submission/components/ProgramMenu.tsx
ciaranschutte: https://github.com/icgc-argo/rdpc-ui/pull/108/commits/5312b370c5a6658f0abbf8f41ff5a26a95a64338

Can't move it out of `render` because it depends on state but I moved it to it's own variable

src/app/(post-login)/submission/components/ProgramMenu.tsx
```diff
@@ -47,7 +47,68 @@ const StatusMenuItem: FC<{ children: ReactNode }> = ({ children }) => {
...
-const parseGQLResp = (data: SideMenuQuery) => {
+const ProgramMenu = ({ searchQuery }: { searchQuery: string }) => {
+	const params = useParams();
+	const pathname = usePathname();
+	const activeProgramName = typeof params.shortName !== 'string' ? '' : params.shortName;
+
+	const { data: programsData, loading, error } = useQuery(SIDEMENU_PROGRAMS);
+	const programs = programsData?.programs || [];
+
+	const [activeProgramIndex, setActiveProgramIndex] = useState(-1);
```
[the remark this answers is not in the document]
ciaranschutte: https://github.com/icgc-argo/rdpc-ui/issues/111
keeping this scoped to page permissions to get it done 
made ticket for this unrelated change

src/app/components/Page.tsx
```diff
@@ -0,0 +1,84 @@
...
+		notFound();
+	} else if (isAuthorized) {
+		return <Component {...urlParams} />;
+	}
+
+	redirect('/403');
+};
+
+// Component type for this project because we use the Emotion compiler
+type Component = (...args: any[]) => EmotionJSX.Element;
```
joneubank, line 73: Cute solution, any chance this will be required in more places with HOCs so maybe we export it from someplace to be reusable?

Also, at a glance, the `any` here could be a generic. I'll try to write out an example of this for you.

src/app/components/Page.tsx
```diff
@@ -0,0 +1,84 @@
...
+// returns Comp or unauthorized based on acceptedRoles
+// params value and types are based on dynamic route of page
+export const pageWithPermissions =
+	// eslint-disable-next-line react/display-name
+	(Component: Component, acceptedRoles: AcceptedRole[]) => (params: any) => {
+		const props = { Component, acceptedRoles, urlParams: params };
+
+		// needs to be Component or hook because we use hooks to check for auth/roles
+		return <Page {...props} />;
+	};
```
joneubank, line 84: Alright I've gone into this more in depth... The actual types this is accepting is URLParams, since this is providing to the page layouts the contents of the page URL params:

```ts
type URLParams = {
    params: {
        [key: string]: string | string[] | undefined;
    } | undefined;
    searchParams: {
        [key: string]: string | string[] | undefined;
    } | undefined;
}
```
Unfortunately, both `params` and `searchParams` are optional (potentially undefined) so when you make the change I am suggesting you will need to update the prop types of all pages that use the `pageWithPermissions` HOC to reflect this URLProps structure and throw errors when the page is accessed.

```suggestion
// Component type for this project because we use the Emotion compiler
type JSXComponent = (args: URLParams) => EmotionJSX.Element;

// returns Comp or unauthorized based on acceptedRoles
// params value and types are based on dynamic route of page
export const pageWithPermissions =
	// eslint-disable-next-line react/display-name
	function (Component: JSXComponent, acceptedRoles: AcceptedRole[]) {
		return function PageWithPermissions(urlParams: URLParams) {
			const props = { Component, acceptedRoles, urlParams };

			// needs to be Component or hook because we use hooks to check for auth/roles
			return <Page {...props} />;
		};
	};
```

src/app/components/Page.tsx
joneubank: Ignore this comment in favour of https://github.com/icgc-argo/rdpc-ui/pull/108/files#r1396014995

src/app/components/Page.tsx
joneubank: I may have another draft of this coming up with a more generic solution that will work with components that don't have the program in the URLParams as well

ciaranschutte:
@joneubank mergeed in your TS / Component changes. added the extra part for undefined ego jwt to be accounted for in SSR prerender

src/app/(post-login)/submission/components/ProgramMenu.tsx
```diff
@@ -47,7 +47,68 @@ const StatusMenuItem: FC<{ children: ReactNode }> = ({ children }) => {
 	);
 };
 
-const parseGQLResp = (data: SideMenuQuery) => {
+const ProgramMenu = ({ searchQuery }: { searchQuery: string }) => {
+	const params = useParams();
+	const pathname = usePathname();
```
joneubank, line 52: I'm not going to request changes for this but...

When we have reliance on param and pathname variables it would probably good to call this out in a TSDoc before the component declaration. It would be best if every component/function had such a documentation block, but especially when there is some usage of URL parameters it would be nice to call out at the top that what the component relies on. This would be a comment that says "Uses query param: `shortName` representing the currently selected program's shortName" and "Uses the URL path to identify which page from the menu is selected. Available pages are `/submission`, `/registration`, and `/clinical-submission`." Or something similar but formatted better and not mixing up options from the multiple components declared in this file.

src/app/components/Page.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+type AcceptedRole = keyof UserRoleList;
+
+// Component type for this project because we use the Emotion compiler
+type JSXComponent<T extends object> = (args: T) => EmotionJSX.Element;
+
+// returns Comp or unauthorized based on acceptedRoles
+// params value and types are based on dynamic route of page
+export const pageWithPermissions =
+	// eslint-disable-next-line react/display-name
```
joneubank, line 37: This linter comment might not be needed now, that's why I was refactoring the returned anonymous function as a named function.

joneubank (approved):
Beauty, thanks.

PR #110, opened 2023-11-14
💧 #1024 Login Hydration Bug

demariadaniel:
Fixes Dev Hydration errors related to Auth Cookies getting out of sync between Server/Client

Additional permissions / user role changes in #1024 are related to changes here: https://github.com/icgc-argo/rdpc-ui/pull/108


src/app/api/config/route.ts
```diff
@@ -17,12 +17,10 @@
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-import { headers } from 'next/headers';
 import { NextResponse } from 'next/server';
 import { getAppConfig } from './config';
 
 export async function GET() {
-	headers();
```
ciaranschutte, line 25: why is this removed? @demariadaniel

demariadaniel: Maybe I misunderstand, but is it not unused? `headers()` returns an object but we do nothing with that object; are you calling this to force the route into a dynamic route? 
I hadn't observed any functional difference but could be wrong

ciaranschutte: Yes it's to force it not to be cached. Runtime var injection in this way will mostly like only be one GET request. But I'm unclear on if this caches at any stage during build time, specifically Nextjs building. unlikely there's a case where this route is HIT with no runtime vars, then ran again with cached GET and some runtime vars...but simply having `headers()` is an easy safeguard to "not cache"

ciaranschutte:
Code looks ok. 
I'm unclear on two things:
1 - is there a need to have this in the `RouteLayout` and prop drill?
2 - Why is this needed? specifically why isn't the functionality in `AuthProvider` working (i'll follow up with specific code line comment)

src/app/layout.tsx
```diff
@@ -46,10 +47,13 @@ async function getAppConfig() {
 
 export default async function RootLayout({ children }: { children: ReactNode }) {
 	const appConfig = await getAppConfig();
+	const jwt = cookies().get(EGO_JWT_KEY)?.toString();
```
ciaranschutte, line 50: https://github.com/icgc-argo/rdpc-ui/pull/110/files#diff-56ef2ffbaaa90be7e518b149c6d2a06a7560f741639aef847658fe858ebfbb2bR60 
similar functionality? One looks like it's using Next apis and the other is native?

I think the cookie return value will either be falsy or a string, so can likely remove `toString()` unless I'm missing something

src/app/api/config/route.ts
demariadaniel: Thanks. Restored w/ comment.

src/app/layout.tsx
justincorrigible, line 50: At the risk of stating what's been mentioned elsewhere (I don't see the comment on this line :shrug:):
anything to do with JWT management should go in an auth context. it doesn't seem to be needed at the layout level.

src/app/hooks/AuthProvider.tsx
```diff
@@ -33,7 +34,6 @@ import {
 	useContext,
 	useState,
 } from 'react';
-import { EGO_JWT_KEY, LOGIN_NONCE } from '../constants';
```
justincorrigible, line 36: :pray:!!!

src/app/App.tsx
```diff
@@ -59,11 +59,19 @@ PortalParent.displayName = 'PortalParent';
...
+}: {
+	children: ReactNode;
+	config: any;
+	jwt?: typeof document.cookie;
+}) => (
 	<ThemeProvider>
 		<QueryClientProvider client={queryClient}>
 			<AppProvider config={config}>
-				<AuthProvider>
+				<AuthProvider jwt={jwt}>
```
justincorrigible, line 74: as a segue to my previous comment (and I see now Ciaran asked about this in his review), it doesn't look like any parents of the `AuthProvider` have any use or need for the jwt, so it would be best to relocate the cookie getter to the context itself :cookie:

demariadaniel: cc @ciaranschutte 
In principle, yes. The problem here is hydration between client + server. See: https://github.com/icgc-argo/rdpc-ui/blob/develop/src/app/App.tsx#L19

We pre-render the page, but there are cookies in the browser. So the pre-render and the browser render don't match, and the AuthProvider does not have a cookie server side, because there is none on the initial server request.

That's why I was using Cookie in layout; App and AuthProvider are 'use client'

justincorrigible: oh! I see! sounds like a good case for having only part of the auth handling be `client` side e.g. the react context, I imagine (my "what's allowed server-side"-fu is not strong enough). 

My concern here is how the auth stuff may start bleeding out of its scope, breaking the separation of concerns principle. Is it possible to have a server-side pre-render (or whatever the right term is) for the auth handlers?

justincorrigible (approved):
Changes are looking much better.
Hopefully this will solve a lot of problems everywhere else 🤞

Squash and merge away!

PR #113, opened 2023-11-16
✨ PageWithPermissions as generic wrapper for any component

joneubank:


src/app/components/Page.tsx
```diff
@@ -21,64 +21,51 @@
...
-	} else if (isAuthorized) {
-		return <Component {...urlParams} />;
-	}
-
-	redirect('/403');
-};
 
 // Component type for this project because we use the Emotion compiler
-type Component = (...args: any[]) => EmotionJSX.Element;
+type JSXComponent<T extends object> = (args: T) => EmotionJSX.Element;
```
ciaranschutte, line 32: Could we use React.ComponentType here?
It covers react component vs a (more specific I think) jsx component

joneubank: probably, give it a try.

ciaranschutte (approved):

PR #114, opened 2023-11-20
📄 Feat/1024 Program Visibility Rules

demariadaniel:
- Filter Program Query by Data Center
- Show/Hide Links (Submission/Registration etc) based on User Role
- Toggle for Program sidebar dropdown menu open/close
- Extra JWT validation in Auth Provider

Covers both: 
https://app.zenhub.com/workspaces/icgc-argo-platform-dk-production-board-5e542d38415f5034e9fed89d/issues/gh/icgc-argo/roadmap/979 
& 
https://app.zenhub.com/workspaces/icgc-argo-platform-dk-production-board-5e542d38415f5034e9fed89d/issues/gh/icgc-argo/roadmap/1024

Admin View
Screenshot 2023-11-23 at 11 45 42 AM

Collaborator View
Screenshot 2023-11-23 at 11 45 08 AM


src/app/(post-login)/submission/components/ProgramList.tsx
```diff
@@ -18,14 +18,16 @@
...
 import { Table, Typography } from '@icgc-argo/uikit';
+import orderBy from 'lodash/orderBy';
 import { columns } from '../program-table/config';
 
-export type ProgramData = any;
 export type ArgoMembershipKey = 'FULL' | 'ASSOCIATE';
 
-export default function ProgramList({ programs }: { programs: ProgramData[] }) {
+export default function ProgramList({ programs }: { programs: Program[] }) {
+	const sortedProgramList = orderBy(programs, 'name');
```
demariadaniel, line 30: If there's only 1 Program per Data Center then sorting won't be necessary...

src/app/(post-login)/submission/components/ProgramMenu.tsx
```diff
@@ -55,26 +66,29 @@ const ProgramMenu = ({ shortNameSearchQuery }: { shortNameSearchQuery: string })
...
-			const url = `/submission/program/${shortName}/registration`;
-			router.push(url);
-		};
+		? programs.filter(notNull)
+		: programs
+				.filter(notNull)
+				.filter((program) => program.shortName.search(new RegExp(shortNameSearchQuery, 'i')) > -1);
+	const sortedProgramList = orderBy(filteredPrograms, 'shortName');
+
+	const [activeProgram, setActiveProgram] = useState(activeProgramName);
```
demariadaniel, line 89: Program dropdown was not closing which is why I've added this

src/app/hooks/AuthProvider.tsx
```diff
@@ -45,6 +45,7 @@ type AuthContextValue = {
 	logOut: () => void;
 	permissions: string[];
 	TokenUtils: ReturnType<typeof createEgoUtils>;
+	userPrograms: string[];
```
demariadaniel, line 48: Added in early stages of PR to filter by logged in user permissions, may no longer be necessary

src/app/(post-login)/submission/program-table/CancerTypes.tsx
```diff
@@ -18,14 +18,17 @@
  */
 'use client';
 
-const CancerTypes = ({ types }: { types: string[] }) => (
+import { Maybe } from '@/__generated__/graphql';
+
+const CancerTypes = ({ types }: { types: Maybe<Maybe<string>[]> | undefined }) => (
```
demariadaniel, line 23: Update Types to match API generated files

src/app/hooks/AuthProvider.tsx
demariadaniel: Removed

src/app/(post-login)/submission/components/ProgramMenu.tsx
ciaranschutte: https://github.com/icgc-argo/rdpc-ui/pull/115/files#diff-b1be0f38440ac03e9ef18232ac95c5f998fe4162a630b28ff29d72cc538e7d1c 

I moved the state to the url in the above PR

I think there's likely a way we can fix the dropdown without reverting back to "local" state ie. `useState`

src/app/(post-login)/submission/components/ProgramMenu.tsx
```diff
@@ -54,27 +65,29 @@ const StatusMenuItem: FC<{ children: ReactNode }> = ({ children }) => {
...
-	const setActiveProgram =
-		(shortName: string): MouseEventHandler =>
-		() => {
-			const url = `/submission/program/${shortName}/registration`;
-			router.push(url);
-		};
+		? programs.filter(notNull)
+		: programs
+				.filter(notNull)
+				.filter((program) => program.shortName.search(new RegExp(shortNameSearchQuery, 'i')) > -1);
```
ciaranschutte, line 85: I'm not sure the `.filter(notNull)` is needed. the gql types are currently broke but this can be investigated

src/app/(post-login)/submission/components/ProgramMenu.tsx
```diff
@@ -172,28 +202,76 @@ const MenuContent = ({ programName }: { programName: string }) => {
...
-							{statusIcon}
-						</StatusMenuItem>
-					}
-					selected={pathnameLastSegment === 'registration'}
-				/>
+			{/** Dashboard */}
+			<Link
+				as={PROGRAM_DASHBOARD_PATH.replace(PROGRAM_SHORT_NAME_PATH, programName)}
+				href={PROGRAM_DASHBOARD_PATH}
+			>
```
ciaranschutte, line 209: why is adding `as` prop needed?

src/app/(post-login)/submission/components/ProgramMenu.tsx
```diff
@@ -172,28 +202,76 @@ const MenuContent = ({ programName }: { programName: string }) => {
...
+				<MenuItem level={3} content="Dashboard" selected={PROGRAM_DASHBOARD_PATH === pathname} />
```
ciaranschutte, line 210: Let's ensure this works as expected. 

```typescript
export const SUBMISSION_PATH = `/submission`;
export const PROGRAM_SHORT_NAME_PATH = `[shortName]`;
export const PROGRAM_DASHBOARD_PATH = `${SUBMISSION_PATH}/program/${PROGRAM_SHORT_NAME_PATH}/dashboard`;
```

so `PROGRAM_DASHBOARD_PATH` will end up as `/submission/program/[shortName]/dashboard`

Nextjs is routing using `[..foo..]` as dynamic values 
this path constant (and any others that interpolate a dynamic path string) will never match an actual `pathName` unless you interpolate the dynamic segment with the path constant you're checking against

ciaranschutte: these constants come in handy for a usage like: 
```javascript
router.push(PROGRAM_DASHBOARD_PATH.replace(PROGRAM_SHORT_NAME_PATH, shortName));
```
but you're always gonna have to get the dynamic route part injected one way or another

src/app/hooks/AuthProvider.tsx
```diff
@@ -68,18 +68,17 @@ export function AuthProvider({ children }: { children: ReactNode }) {
...
 			setEgoJwt(storedToken);
 		}
 		setAuthLoading(false);
-	}, [storedToken]);
-
-	const TokenUtils = createEgoUtils(config.EGO_PUBLIC_KEY);
-	const router = useRouter();
-	const loginStateOnPageLoad = !egoJwt?.length;
-	const [authLoading, setAuthLoading] = useState(loginStateOnPageLoad);
+	}, [storedToken, TokenUtils]);
```
ciaranschutte, line 81: I'm unclear on why `TokenUtils` is a dependency?

src/app/hooks/AuthProvider.tsx
```diff
@@ -97,7 +96,8 @@ export function AuthProvider({ children }: { children: ReactNode }) {
 		setAuthLoading(false);
 	};
 
-	const permissions = TokenUtils.getPermissionsFromToken(egoJwt);
+	const permissions =
+		(TokenUtils.isValidJwt(egoJwt) && TokenUtils.getPermissionsFromToken(egoJwt)) || [];
```
ciaranschutte, line 100: I think `TokenUtils.isValidJwt` is already checked when you set `egoJwt` state.

src/app/hooks/AuthProvider.tsx
demariadaniel: Linter. I think `TokenUtils.isValidJwt(storedToken)` would make more sense.

src/app/(post-login)/submission/components/ProgramMenu.tsx
demariadaniel: This is direct from Platform, I can look into it

ciaranschutte: Gotcha. So the `as` is usually when you have middleware of some kind. You want the functionality to be to specific urls, but need it displayed as single. The example they give in their docs is something like `/auth/dashboard` vs `/dashboard` but you still want the link as `/dashboard`

src/app/(post-login)/submission/components/ProgramMenu.tsx
```diff
@@ -160,6 +183,13 @@ const MenuContent = ({ programName }: { programName: string }) => {
 		},
 	});
 
+	const { isRDPCAdmin, isDCCAdmin, isProgramAdmin, isDataSubmitter, isCollaborator } = userRoles;
+
+	const userCanSubmitData = isDataSubmitter || isRDPCAdmin || isDCCAdmin || isProgramAdmin;
+	const userCanViewData =
+		isCollaborator || isDataSubmitter || isRDPCAdmin || isDCCAdmin || isProgramAdmin;
+	const userCanManageProgram = isDataSubmitter || isRDPCAdmin || isDCCAdmin || isProgramAdmin;
+
```
ciaranschutte, line 192: What are your thoughts on exposing more specific functionality from within `useUserRoles`?
if we're just checking can user view data... can we just check for that specific thing? vs checking if a user is part of any role that has that permission. Does that make sense? These role booleans are just built from ego token checks. So here we're checking a bunch of supersets instead of just checking for the permission.

demariadaniel: We could check `${PROGRAM}.READ` / `${PROGRAM}.WRITE` permissions, which is how I approached this branch originally. See:
https://github.com/icgc-argo/rdpc-ui/pull/114/commits/d3b5622f7f958385b35e58165390ff5e5fbc49b3
(original code that I removed)

These booleans came from matching useUserRoles to the outlines of which roles should see which page in ticket #979:
https://app.zenhub.com/workspaces/icgc-argo-platform-dk-production-board-5e542d38415f5034e9fed89d/issues/gh/icgc-argo/roadmap/979

src/app/(post-login)/submission/components/ProgramMenu.tsx
demariadaniel: I will remove `as` for now and update `href` to `href={PROGRAM_MANAGE_PATH.replace(PROGRAM_SHORT_NAME_PATH, programName)}` for each path

demariadaniel: This is rendering correctly

src/app/(post-login)/submission/components/ProgramMenu.tsx
demariadaniel: see above discussion, updated

src/app/(post-login)/submission/components/ProgramMenu.tsx
demariadaniel: Discussing on Slack; there needs to be separation between opening/closing menu items and navigating to a specific page

src/app/(post-login)/submission/components/ProgramMenu.tsx
ciaranschutte: potentially just change line 81 to :
```javascript
	const programs = programsData?.programs?.filter(notNull) || [];
```

ciaranschutte: as we're filtering by `notNull` either way

src/app/hooks/AuthProvider.tsx
```diff
@@ -68,18 +68,17 @@ export function AuthProvider({ children }: { children: ReactNode }) {
 	const config = useAppConfigContext();
 	const storedToken = getStoredToken();
 	const [egoJwt, setEgoJwt] = useState('');
+	const TokenUtils = createEgoUtils(config.EGO_PUBLIC_KEY);
+	const router = useRouter();
+	const loginStateOnPageLoad = !egoJwt?.length;
```
joneubank, line 73: very confusing variable name... loginState i would expect a state as a string from list of options... this is just a bool... could this just be `isLoggedIn = !!egoJwt`?

src/app/hooks/AuthProvider.tsx
```diff
@@ -68,18 +68,17 @@ export function AuthProvider({ children }: { children: ReactNode }) {
 	const config = useAppConfigContext();
 	const storedToken = getStoredToken();
 	const [egoJwt, setEgoJwt] = useState('');
```
joneubank, line 70: You've inheritted this line but...
```suggestion
	const [egoJwt, setEgoJwt] = useState<string | undefined>(undefined);
```

Best to use undefined for missing values.

src/app/hooks/AuthProvider.tsx
```diff
@@ -97,7 +96,7 @@ export function AuthProvider({ children }: { children: ReactNode }) {
 		setAuthLoading(false);
 	};
 
-	const permissions = TokenUtils.getPermissionsFromToken(egoJwt);
+	const permissions = TokenUtils.getPermissionsFromToken(egoJwt) || [];
```
joneubank, line 99: `getPermissionsFromToken` always returns `string[]` so this `|| []` is not needed.

https://github.com/icgc-argo/ego-token-utils/blob/develop/src/index.ts#L85

src/app/(post-login)/submission/program-table/CancerTypes.tsx
ciaranschutte: Would it be better to `pick` the sub type of `cancerTypes` from here: https://github.com/icgc-argo/rdpc-ui/blob/develop/src/__generated__/graphql.ts#L2744

`Maybe` is just a utility type being imported from the generated types - so we may as well use the full generated type?

src/app/(post-login)/submission/program-table/DonorStatus.tsx
```diff
@@ -18,15 +18,17 @@
...
 
-type DonorStatusProps = { submittedDonors: number; commitmentDonors: number };
+type DonorStatusProps = { submittedDonors?: Maybe<number>; commitmentDonors?: Maybe<number> };
 
 const DonorStatus = ({ submittedDonors, commitmentDonors }: DonorStatusProps) => {
 	const theme = useTheme();
-	const numerator = submittedDonors < 0 ? 0 : submittedDonors;
-	const denominator = commitmentDonors;
+	const numerator =
+		!(typeof submittedDonors === 'number') || submittedDonors < 0 ? 0 : submittedDonors;
```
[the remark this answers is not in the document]
demariadaniel: Updated to `submittedDonors && submittedDonors > 0 ? submittedDonors : 0`

src/app/hooks/AuthProvider.tsx
demariadaniel: This has a few impacts on other type definitions so I've updated `permissions` like this:
`const permissions = egoJwt ? TokenUtils.getPermissionsFromToken(egoJwt) : [];`

src/app/(post-login)/submission/program-table/CancerTypes.tsx
demariadaniel: In principal yes but in practice `Pick<ProgramsListQuery['programs'], 'cancerTypes'>` where both `programs` and `cancerTypes` are potentially undefined becomes tricky to untangle

src/app/(post-login)/submission/components/ProgramMenu.tsx
demariadaniel: Nice, much improved

src/app/(post-login)/submission/components/ProgramMenu.tsx
```diff
@@ -55,19 +66,26 @@ const ProgramMenu = ({ shortNameSearchQuery }: { shortNameSearchQuery: string })
...
 	const filteredPrograms = !shortNameSearchQuery.length
-		? programs
-		: programs.filter(
-				(program) =>
-					program && program.shortName.search(new RegExp(shortNameSearchQuery, 'i')) > -1,
-		  );
+		? programs.filter(notNull)
+		: programs
+				.filter(notNull)
+				.filter((program) => program.shortName.search(new RegExp(shortNameSearchQuery, 'i')) > -1);
```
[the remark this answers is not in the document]
demariadaniel: See Ciaran's suggestion above, this whole thing was able to be simplified into one definition

ciaranschutte:
@demariadaniel not sure how far you've gone with the side menu bug. Thoughts on leaving it out of this PR and we can tackle it as a separate ticket/PR and keep this one for the program visibility rules? Hopefully quicker to get this merged through that way.

.env
```diff
@@ -21,3 +21,5 @@ NEXT_PUBLIC_GATEWAY_API_ROOT=
 NEXT_PUBLIC_RECAPTCHA_SITE_KEY=
 
 NEXT_PUBLIC_RUNTIME_CONFIG_URL=http://localhost:3000/api/config
+
+NEXT_PUBLIC_DATA_CENTER='DC1'
```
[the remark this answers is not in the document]
demariadaniel: Yes, updated. This was one thing about the new Next.js approach that I had questions about.

src/app/(post-login)/submission/components/ProgramMenu.tsx
```diff
@@ -92,24 +112,28 @@ const ProgramMenu = ({ shortNameSearchQuery }: { shortNameSearchQuery: string })
 					<MenuItem
 						level={2}
 						content="All Programs"
-						onClick={setActiveProgram('')}
+						onClick={() => {
+							setActiveProgram('');
+						}}
```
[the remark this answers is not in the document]
demariadaniel: Np updated

demariadaniel:
> @demariadaniel not sure how far you've gone with the side menu bug. Thoughts on leaving it out of this PR and we can tackle it as a separate ticket/PR and keep this one for the program visibility rules? Hopefully quicker to get this merged through that way.

Yes, that can be revisited
PR feedback has been addressed!

src/app/(post-login)/submission/components/ProgramMenu.tsx
```diff
@@ -71,7 +71,7 @@ const ProgramMenu = ({ shortNameSearchQuery }: { shortNameSearchQuery: string })
 	const { DATA_CENTER } = useAppConfigContext();
 
 	// params can be arrays from dynamic routing
-	const activeProgramName = typeof params.shortName === 'string' ? params.shortName : '';
+	let activeProgramName = typeof params.shortName === 'string' ? params.shortName : '';
```
ciaranschutte, line 74: `const` vs `let` in react world is a code smell to me because there's not a whole lot of uses cases for `let` in react.
using `let` here implies we want this value to change which implies it's state and UI reacting to said state.

because React will batch and reconcile state changes internally to write to it's virtual DOM we have no control over it if we use let, const .. var.. in short. React might know about the change - but it might not. I don't know if React just ignores "state like variables".

because we're changing the `activeProgramName` in a click handler to change a `let`, we haven't told React it's happening. we should use React.useState

for the sake of covering all the bases here - there are still cases to use regular variables! derived values are a great example. here I'm specifically commenting on this because we are using it as state that pieces of UI depend on

@joneubank @justincorrigible  have some more experience than me on this and may be able to explain it better ( I could also be straight up incorrect ;) )

Following on from my earlier comment I think we should just leave this activeProgram side menu bug for it's own isolated ticket/PR. Can we revert the changes to the sidemenu in this PR so it's not merged in with the other work?

src/lib/gql.ts
```diff
@@ -41,7 +41,7 @@ const createInMemoryCache = () =>
 
 type Config = {
 	gateway: string;
-	jwt: string;
+	jwt: string | undefined;
```
ciaranschutte, line 44: I'm unclear why this can now be `undefined`?
The type is not exported (not used anywhere else) and there's no other changes to this file or the ApolloProvider which is a parent of AuthProvider

src/app/(post-login)/submission/components/ProgramMenu.tsx
```diff
@@ -168,6 +169,36 @@ const parseProgramStatusGQLResp = (data: SideMenuProgramStatusQuery | undefined)
 	};
 };
 
+const renderDataSubmissionLinks = (
+	programName: string,
+	statusIcon: EmotionJSX.Element | null,
```
ciaranschutte, line 180: can `React.ReactNode` be used here? I think we should avoid depending on EmotionJSX if possible. Their typings aren't always inline with Reacts typings.

src/app/(post-login)/submission/program-table/DonorStatus.tsx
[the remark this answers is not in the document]
justincorrigible: good change... possibly doesn't need the first `submitterDonors &&` at that point.

if submitterDonnors is falsy, it'll never be greater than 0

src/lib/gql.ts
justincorrigible: that's likely because of this?
https://github.com/icgc-argo/rdpc-ui/pull/114#discussion_r1404368017

src/app/(post-login)/submission/components/ProgramMenu.tsx
justincorrigible: You're right, Emotion coupling is something we'd want to keep to a minimum; and considering the type you suggest would allow strings, etc. I think more than a node, an icon needs to be an element.

demariadaniel: Yes thankfully we can. I used Emotion as it was the suggested type in VSCode

Screenshot 2023-11-28 at 10 27 10 AM

src/lib/gql.ts
demariadaniel: Justin is correct, due to updating `egoJwt` definition, createApploClient args (in `ApolloProvider`) are invalid without this change

src/app/(post-login)/submission/components/ProgramMenu.tsx
demariadaniel: You're correct and I agree, this is merely a side effect of not fixing the Side Menu open/close bug as part of this ticket, which would require introducing component state in addition to the URL state.

I will use `const activeProgramName` and the bug remains for now, and fix the rest in a separate branch.

src/app/(post-login)/submission/program-table/DonorStatus.tsx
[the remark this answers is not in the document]
demariadaniel: @justincorrigible No `Boolean(-1) = true`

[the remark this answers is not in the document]
demariadaniel: So this can just be `!!submittedDonors` then

If submittedDonors < 0 we have a very different problem lol

src/app/(post-login)/submission/program-table/Countries.tsx
```diff
@@ -18,14 +18,17 @@
...
+const Countries = ({ countries }: { countries: Maybe<Maybe<string>[]> | undefined }) => (
 	<div>
-		{countries.map((country, i) => (
-			<div key={country}>
-				{country}
-				{i < countries.length - 1 && ','}
-			</div>
-		))}
+		{countries &&
+			countries.map((country, i) => (
```
justincorrigible, line 26: If you're trying to account for countries being defined, this is a cleaner approach since ES2020.
and in this case, the Optional Chaining operator also checks that map is a method of countries, for good measure...
hard to tell with that nested Maybe whether it'd be useful or even necessary.
```suggestion
		{countries?.map?.((country, i) => (
```

src/app/(post-login)/submission/program-table/CancerTypes.tsx
```diff
@@ -18,14 +18,17 @@
...
 	<div>
-		{types.map((cancerType, i) => (
-			<div key={cancerType}>
-				{cancerType}
-				{i < types.length - 1 && ','}
-			</div>
-		))}
+		{types &&
+			types.map((cancerType, i) => (
```
justincorrigible, line 26: Optional chaining is a good approach here too
```suggestion
		{types?.map?.((cancerType, i) => (
```

demariadaniel: Shrink 'em down, updated

src/app/(post-login)/submission/program-table/CancerTypes.tsx
joneubank: There's two options here:
1. Use the generated types from the GQL Schema:
```ts
import { Program } from '@/__generated__/graphql';

const CancerTypes = ({ types }: { types: Program['cancerTypes'] }) => (
```
This will ensure that your prop type can be read directly from a query on that generated type. The downside here is that if the GQL schema changes then you need to change this React Component.

2. Define the interface to accept only the type you want:

```ts
const CancerTypes = ({ types }: { types: string[] }) => (
```
and perform type filtering before passing data to your component:

```ts
			<CancerTypes
				types={
					cancerTypes
						? cancerTypes.filter((value): value is string => typeof value === 'string')
						: []
				}
			/>
```
This has the advantage of performing your type filtering in a location that is already aware that the data comes from GQL, and now your component never needs to care about where the data comes from: it is assured to get just the type it wants.

src/app/(post-login)/submission/components/ProgramMenu.tsx
ciaranschutte: `<MenuItem level={3} content="Dashboard" selected={PROGRAM_DASHBOARD_PATH === pathname} />`
 
this `selected` prop for example will be `selected={'/submission/program/[shortName]/dashboard' === '/submission/program/CIA-IE/dashboard'`

`PROGRAM_DASHBOARD_PATH` will always have the `[shortName]` dynamic placeholder in the string until you substitute it dynamically with the actual pathname of a page.

this is introducing a bug.

src/app/(post-login)/submission/program-table/CancerTypes.tsx
demariadaniel: Split the difference: 
config.tsx: `<CancerTypes types={cancerTypes || []} />`
CancerTypes.tsx: `const CancerTypes = ({ types }: { types: Program['cancerTypes'] | [] }) => (...)`

joneubank: Option 2 recommended. Keep the GQL knowledge limited to a few places, doesn't need to be in your display components.

joneubank:
```ts
<CancerTypes types={cancerTypes || []} />
```
would be compatible with the type signature:
```ts
({ types }: { types: Program['cancerTypes']}); // no `| []` needed
```

The `| []` could cause you more problems. I wouldn't go that route.

demariadaniel: Would this work?
`<CancerTypes types={cancerTypes?.filter(notNull) || []} />`
`const CancerTypes = ({ types }: { types: string[] }) => (...)`

joneubank: Yeah that's basically a condensed form of what I wrote. I like it.

src/app/(post-login)/submission/components/ProgramMenu.tsx
demariadaniel: Updated in: https://github.com/icgc-argo/rdpc-ui/pull/114/commits/afc07b6f8dcaba6b6b5abb6fa37ee6c2dc80fde3

.env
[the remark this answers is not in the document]
ciaranschutte: they break convention.
.env is now for defaults and "should be commmitted"

src/app/(post-login)/submission/components/ProgramMenu.tsx
```diff
@@ -259,18 +269,18 @@ const MenuContent = ({
...
-				<Link href={PROGRAM_MANAGE_PATH.replace(PROGRAM_SHORT_NAME_PATH, programName)}>
+				<Link href={getProgramPath(PROGRAM_MANAGE_PATH)}>
 					<MenuItem
 						level={3}
 						content="Manage Program"
-						selected={PROGRAM_MANAGE_PATH === pathname}
+						selected={pathnameLastSegment === 'manage'}
 					/>
 				</Link>
 			)}
```
ciaranschutte, line 286: This is fine. your original `replace` in the `href` prop was valid. 

It was just the `selected` prop part that needed to have the dynamic part injected which is now not needed because it's been reverted to how it was before.

ciaranschutte (approved):
after merge conflict res this is ok to merge

PR #118, opened 2023-11-29
🪱 Bug: Auth Destructuring Error

demariadaniel:
Addresses issue w/ `Cannot destructure property 'auth' of 'e' as it is undefined` error in browser


src/lib/gql.ts
```diff
@@ -55,7 +55,7 @@ export const createApolloClient = (config: Config) => {
 	const authLink = new ApolloLink((operation, forward) => {
 		operation.setContext(({ headers }: { headers: any }) => ({
 			headers: {
-				authorization: `Bearer ${jwt}`,
+				authorization: jwt ? `Bearer ${jwt}` : '',
```
joneubank, line 58: Perhaps we shouldnt be setting an authorization header if we have no token?

src/app/hooks/ApolloProvider.tsx
```diff
@@ -29,7 +29,7 @@ export const ApolloProvider = ({ children }: { children: ReactNode }) => {
 	const auth = useAuthContext();
 	const { GATEWAY_API_ROOT } = useAppConfigContext();
 	const config = {
-		jwt: auth.egoJwt,
+		jwt: auth?.egoJwt,
```
joneubank, line 32: `useAuthContext ` should never return as undefined, the type is well defined and provided with a good default value. This `?` shouldn't be necessary.

src/lib/gql.ts
demariadaniel: Sure

```
	const authHeader = jwt ? {authorization: `Bearer ${jwt}`} : {};	

	const authLink = new ApolloLink((operation, forward) => {
		operation.setContext(({ headers }: { headers: any }) => ({
			headers: {
				...authHeader,
				...headers,
			},
```

src/app/hooks/ApolloProvider.tsx
demariadaniel: OK fair this was a misfire, this is the only key `auth` in our app, but the error is key `auth` in a dependency
Reverted, but don't think the Headers issue above is enough to explain the bug either... investigation continues

src/lib/gql.ts
justincorrigible: Not asking for change: this has lower cognitive burden.
```
const authHeader = jwt && {authorization: `Bearer ${jwt}\`};
```

joneubank: Doesn't that break the next section's `...authHeader` spread?

justincorrigible: nup. try it out ;) 

`{ ...false } == { }`

demariadaniel: Weird!

PR #119, opened 2023-11-29
🕷️  Default Values for Run time Config URL + Data Center

demariadaniel:
Protect against empty values for Config URLs (source of bug in Dev)


src/app/api/config/config.ts
```diff
@@ -66,7 +66,7 @@ export const getAppConfig = (serverEnv: any): AppConfig => {
 			serverEnv.NEXT_PUBLIC_EGO_PUBLIC_KEY || process.env.NEXT_PUBLIC_EGO_PUBLIC_KEY || '',
 		UI_VERSION: packageJSON.version,
 		REGION: serverEnv.NEXT_PUBLIC_REGION || process.env.NEXT_PUBLIC_REGION || '',
-		DATA_CENTER: serverEnv.NEXT_PUBLIC_DATA_CENTER || process.env.NEXT_PUBLIC_DATA_CENTER || '',
+		DATA_CENTER: serverEnv.NEXT_PUBLIC_DATA_CENTER || process.env.NEXT_PUBLIC_DATA_CENTER || 'DC1',
```
joneubank, line 69: What is `DC1` ?

joneubank (approved):
Thanks for putting this in place... the values you provided look like placeholders to make dev work, not actual values we want to default to.

It may be time to protect our config files correctly, where we throw an error on startup if required env variables are not provided.

src/app/api/config/config.ts
demariadaniel: DC1 is the only Data Center value we have at the moment

demariadaniel:
> Thanks for putting this in place... the values you provided look like placeholders to make dev work, not actual values we want to default to.
> 
> It may be time to protect our config files correctly, where we throw an error on startup if required env variables are not provided.

@joneubank good point, I guess this is useful in the interim (since we are still testing and working on Dev/QA) but is worthy of some dedicated follow up

PR #125, opened 2023-11-30
Use single PageHeader component consistently 

ciaranschutte:
- same html structure styling with imports used in multiple places
- use single component


src/app/(post-login)/submission/program/page.tsx
```diff
@@ -38,5 +41,20 @@ export default function Submission() {
 	if (loading) return <Loader />;
 	if (error) notFound();
 
-	return <ProgramList programs={programs} />;
+	const canCreate = true;
```
ciaranschutte, line 44: there is logic to this rendering, just not part of this reusable component PR

src/app/components/PageHeader/PageHeader.tsx
```diff
@@ -43,42 +69,16 @@ const ContentHeader: FunctionComponent<ContentHeaderProps> = ({
 				align-items: center;
 				background-color: ${theme.colors.white};
 			`}
+			className={className}
```
justincorrigible, line 72: Not necessary for this PR, as nobody is "reusing" this: I've gotten a habit of using the `classnames` package for reusable components, injecting a `csx('PageHeader', className)` pattern, so it has a default local name, aside from the ones passed in (useful when someone wants to target an element through CSS)

PR #128, opened 2023-12-04
Feat/#55

ciaranschutte:
- `platform-ui` code integrated into `rdpc-ui`
- original code has passed code review 
- changes made from original:
- use new `usePageContext` hook, Next13 doesn't inject page `ctx` anymore
- update `useAuthContext` references
- multiple states cleaned up, or moved to the edge of render tree (ie. not needed in top components, so prop drilling is just noise)
- removed some `useEffect` calls because the value is just a computed value


ciaranschutte:
scrap improvements

PR #130, opened 2023-12-05
❗ Protect Program Side Menu against Falsy Values

demariadaniel:


src/app/(post-login)/submission/components/ProgramMenu.tsx
```diff
@@ -198,10 +198,10 @@ const renderDataSubmissionLinks = (
...
 					<StatusMenuItem>
 						Submit Clinical Data{' '}
-						{programStatusData &&
+						{!!programStatusData &&
 							renderSubmissionStatusIcon(
-								programStatusData.clinicalSubmissionState || null,
-								programStatusData.clinicalSubmissionHasSchemaErrors,
+								programStatusData?.clinicalSubmissionState || null,
+								programStatusData?.clinicalSubmissionHasSchemaErrors || false,
 								isSubmissionSystemDisabled,
```
ciaranschutte, line 205: I'm not seeing how this solves the issue. 
the `!!` isn't needed because it's already a boolean context
the optional chain op isn't needed because default `clinicalSubmissionState` is already `undefined` so this should return `null` and `clinicalSubmissionHasSchemaErrors` default is `false` already.

ciaranschutte:
there's also nothing rendering - so the conditional is working - i'll slack you

PR #137, opened 2023-12-13
adds 2nd gql endpoint config

ciaranschutte:
uses query context to target a specific endpoint
- simplest way without getting into schema stitching
- codegen gql types breaks because it works with one url exclusively
- use a hook and export gql query + api endpoint from files (using a common hook means we can enforce usage, rather than having people needing to remember to specific `context: {apiName: <var>}`


src/app/hooks/useApolloQuery.tsx
```diff
@@ -16,13 +16,24 @@
...
-export default CLINICAL_SCHEMA_VERSION_QUERY;
+export const useApolloQuery = <TData, TVariables>(
+	query: {
+		gql: DocumentNode | TypedDocumentNode<TData, TVariables>;
+		api: keyof typeof apiName;
+	},
+	options?: QueryHookOptions<any, OperationVariables> | undefined,
+): QueryResult<TData, TVariables> => {
+	const mergedOptions = { context: { apiName: query.api }, ...options };
+	// @ts-expect-error
```
ciaranschutte, line 37: @joneubank if you fancy ;) 
TS doesn't like that I'm spreading options into a new object. There's a `NoInfer` types from apollo in use, I can't understand it.

src/app/gql/CLINICAL_SCHEMA_VERSION.ts
```diff
@@ -18,11 +18,15 @@
...
 const CLINICAL_SCHEMA_VERSION_QUERY = gql(`
   query ClinicalSchemaVersion {
     clinicalSubmissionSchemaVersion
   }
 `);
 
-export default CLINICAL_SCHEMA_VERSION_QUERY;
+export const ClinicalSchemaVersionQuery = {
+	gql: CLINICAL_SCHEMA_VERSION_QUERY,
+	api: apiName.clinical,
```
demariadaniel, line 31: OK so just confirming, going forward all queries will need to specify which API they point at?

Should we make Clinical the default? Could potentially reduce some extra lines of code

ciaranschutte: Yes that is correct
We can add a default, but it's only an extra line of code. I don't mind being a little bit more explicit. Especially if it makes it clear

justincorrigible: While i don't hold a strong opinion, and do see the benefit of having a default value, intuitively I'm inclined towards advocating for having the single "Query" helper be agnostic, and then have two separate wrappers that each passes which endpoint you intend on calling. e.g. endpoint1QueryFn and endpoint2QueryFn, both based on baseQueryHelperFn.

demariadaniel: I do like wrapperA vs wrapperB pattern

My original thinking is that Gateway is only needed for one query (currently... could change)
So if majority of queries are all pointing to Clinical, you shouldn't have to specify that for every query
Why change all queries when only 1 needs to be different

justincorrigible: that's a fair point. I'm just trying to prevent the "you have to know this thing that is abstracted elsewhere" that is so on brand with our previous apps, which makes on-boarding so painful.

demariadaniel: Yeah exactly, I think wrapper is the best approach for that reason!

ciaranschutte:
@justincorrigible @demariadaniel refactored to use specific wrapper functions instead of passing params.

demariadaniel (approved):

PR #145, opened 2024-01-12
Table fixes

ciaranschutte:
- fixes the css inline instead of fixing UIKit 
  - some discussion to be had on if we're continuing to support it
  - the ideal fix for this is changing how many things work in UIKit table
- reference `#8` in this table https://wiki.oicr.on.ca/display/icgcargotech/React+Table+v8+Upgrade+Plans


ciaranschutte:
closing in favour of fixing the root causes in UIKit

PR #146, opened 2024-01-12
 change gql mutation uploads to use fetch

ciaranschutte:
* new functionality
- change gql mutation uploads to use fetch
- puts reusable stuff in places where it can be reused


src/app/(post-login)/submission/program/[shortName]/sample-registration/page.tsx
```diff
@@ -96,33 +101,29 @@ const Register = ({ shortName }: { shortName: string }) => {
 	const registrationId = get(clinicalRegistration, 'id', '') || '';
 
 	// handlers
-	const [uploadFile, { loading: isUploading }] = useMutation(
-		UPLOAD_REGISTRATION_MUTATION,
-
+	const uploadFile = useMutation(
+		(formData) => {
+			const url = urlJoin(CLINICAL_API_ROOT, getProgramPath(UPLOAD_REGISTRATION, shortName));
+			return uploadFileRequest(url, formData, egoJwt);
```
Comment (joneubank) [inline src/app/(post-login)/submission/program/[shortName]/sample-registration/page.tsx:107]: This function just returns the promise from the `fetch` to upload the form... is this awaited properly by the `useMutation` hook? There is also no try/catch or other error handling so I assume you tested this and the `useMutation` passes a failed request onto the `onError` function below. It is concerning that we aren't communicating any errors other than "unknown".

joneubank (approved):

src/app/(post-login)/submission/program/[shortName]/sample-registration/page.tsx
Comment (ciaranschutte) [reply src/app/(post-login)/submission/program/[shortName]/sample-registration/page.tsx:107]: the error handling is in the `onError` options property below.
I agree it's concerning. This is the only error case handled currently in `platform-ui`

Comment (ciaranschutte) [reply src/app/(post-login)/submission/program/[shortName]/sample-registration/page.tsx:107]: it's a different way of doing it. this is how `react-query` is does it. "hooks" way

PR #147, opened 2024-01-16
Table fix

ciaranschutte:
- custom TD and TH
- sorting
- sticky cell
related uikit PR https://github.com/icgc-argo/uikit/pull/141


src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable.tsx
```diff
@@ -475,10 +475,12 @@ const ClinicalEntityDataTable = ({
 			.sort(sortEntityData);
 	}
 
+	const styleThickBorderString = `3px solid ${theme.colors.grey}`;
 	const getHeaderBorder = (key) =>
 		(showCompletionStats && key === completionColumnHeaders.followUps) ||
-		(!showCompletionStats && key === 'donor_id')
-			? `3px solid ${theme.colors.grey}`
+		(!showCompletionStats && key === 'donor_id') ||
+		key === 'FO'
```
Comment (justincorrigible) [inline src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable.tsx:482]: what's `FO` in this context? may be useful to abstract this into a declaratively named constant, to facilitate reading.

src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable.tsx
```diff
@@ -636,71 +638,159 @@ const ClinicalEntityDataTable = ({
...
+									`box-shadow: inset 0 ${isSorted === 'asc' ? '' : '-'}3px 0 0 rgb(7 116 211)`};
+								`}
+							>
+								{val}
+							</th>
+						);
+					},
 					maxWidth: noTableData ? 50 : 250,
 					style: noTableData ? noDataCellStyle : {},
+					meta: { customCell: true, customHeader: true },
```
Comment (justincorrigible) [inline src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable.tsx:740]: cool! hadn't seen this property in the react-table docs. or is it a uikit thing perhaps?

src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable.tsx
```diff
@@ -776,6 +867,9 @@ const ClinicalEntityDataTable = ({
 				withSideBorders
 				withPagination
 				showPageSizeOptions
+				withStripes
+				enableColumnResizing={false}
+				enableSorting
```
Comment (justincorrigible) [inline src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable.tsx:872]: in contrast, as follow up to my earlier "alphabetise all the things" comment, TS does a much better job at identifying and pointing out duplicate React props, so I don't care as much here.

...I'd only advocate for it here for consistency and "coding discipline", but I realise other devs don't feel as strongly about that as I do.

src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable.tsx
```diff
@@ -636,71 +638,159 @@ const ClinicalEntityDataTable = ({
...
+							</div>
+						</th>
+					);
+				},
 				headerStyle: completionHeaderStyle,
 				columns: columns.slice(0, 7).map((column) => ({
 					...column,
+					sortingFn: sortEntityData,
+					header: (props) => {
+						const val = props.header.id;
```
Comment (justincorrigible) [inline src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable.tsx:711]: `val` -> `value`, no need to skimp on letters; unless `value` is already in scope, in which case `aValue` vs `bValue` would be better than variants of the same single word.

src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable.tsx
```diff
@@ -636,71 +638,159 @@ const ClinicalEntityDataTable = ({
...
+												margin-right: 6px;
+											`}
+										>
+											For clinical completeness, each donor requires: <br />
+											DO: at least one Donor record <br />
+											PD: at least one Primary Diagnosis record <br />
+											NS: all the registered Normal DNA Specimen record <br />
+											TS: all the registered Tumour DNA Specimen record <br />
+											TR: at least one Treatment record <br />
+											FO: at least one Follow Up record <br />
```
Comment (justincorrigible) [inline src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable.tsx:696]: not a request for change, because I know this is an established pattern and not something just from this PR:
🤔 this dictionary right here is what I wish we do with VS' intellisense, that if you hover over a constant, it also gives you a description of what it means, apart from its string value.
wonder if there's a plugin to do that sort of thing.

Comment (justincorrigible) [reply src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable.tsx:696]: ahh JSDoc to the rescue! no plugin needed 👍 
image

Comment (ciaranschutte) [reply src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable.tsx:696]: nice!

src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable.tsx
```diff
@@ -636,71 +638,159 @@ const ClinicalEntityDataTable = ({
...
+						</th>
+					);
+				},
 				headerStyle: completionHeaderStyle,
 				columns: columns.slice(0, 7).map((column) => ({
 					...column,
+					sortingFn: sortEntityData,
+					header: (props) => {
+						const val = props.header.id;
+						const isThickBorder = val === 'FO';
```
Comment (joneubank) [inline src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable.tsx:712]: I assume this one is thick border cause its the last element? Should the logic be based on it being the last element not a specific element?

Comment (ciaranschutte) [reply src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable.tsx:712]: It's somewhat unclear in regards to the groupings in the table.
Last element requires us to target the last element of the row.

We can't access the row element.
So we'd have to do some logic with the indices from the internal react table data which we shouldn't be touching at this level. That is way more flaky than just hard coding a value. 

DX wise `val === "FO` is very clear. It works and is very much in line with the code quality in the rest of this file

src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable.tsx
Comment (ciaranschutte) [reply src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable.tsx:482]: There's logic like this all over this file.

src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable.tsx
Comment (joneubank) [reply src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable.tsx]: Sounds fine, it is clear whats happening. Just to complete my though there though, do we not have access the index form the .map? partial code example:

```ts
const coreCompletionColumnsCount = 7;
columns: columns.slice(0, coreCompletionColumnsCount).map((column, index) => ({
	...column,
	sortingFn: sortEntityData,
	header: (props) => {
		const isLastElement = index === coreCompletionColumnsCount -1;

```

joneubank (approved):

PR #148, opened 2024-01-17
add clean command

ciaranschutte:
- case with local dev packages, next can cache too much causing issues in modules

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
     "test": "echo 'Add tests please :)'",
     "gql:compile": "graphql-codegen",
     "gql:watch": "graphql-codegen -w",
-    "type-check": "tsc -p ./tsconfig.json"
+    "type-check": "tsc -p ./tsconfig.json",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -11,7 +11,8 @@
     "test": "echo 'Add tests please :)'",
     "gql:compile": "graphql-codegen",
     "gql:watch": "graphql-codegen -w",
-    "type-check": "tsc -p ./tsconfig.json"
+    "type-check": "tsc -p ./tsconfig.json",
+    "clean": "rimraf .next"
```
justincorrigible, line 15: nice! I did something similar in dms-ui, and this here is the end result of my troubleshooting
https://github.com/overture-stack/dms-ui/blob/develop/package.json#L8

ciaranschutte: good shout. added.

justincorrigible: oh I meant "and 'dev'" too, for good measure

PR #151, opened 2024-01-26
Feat/dashboard

ciaranschutte:
- Dashboard page
  - StatsBar comp


src/app/hooks/useTimeout.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const useTimeout = (msTimeout: number = 30000) => {
+	const [isTimeOut, setIsTimeOut] = useState(false);
+	useEffect(() => {
+		const timer = setTimeout(() => {
+			setIsTimeOut(true);
+		}, msTimeout);
+		return () => clearTimeout(timer);
+	}, []);
+	return isTimeOut;
+};
```
joneubank, line 31: The arbitrary 30seconds here is surprising to see. Do we use this default anywhere?

If this is keeping the default value I would add TSDocs to explicitly call it out. 

If I was writing this today, I would not put a default in the function. If we wanted a common timeout value for many instances of this that are using it for a similar operation, then we could reference a global constant that can be modified through config. Use that in each instance instead of setting a default value here.

joneubank (commented):
So much React here, I only skimmed it since its being brought over from the existing UI.

Put a handful of thoughts on things that caught my eye, nothing urgent to fix.

src/app/(post-login)/submission/program/[shortName]/dashboard/components/DonorDataSummary/types.tsx
```diff
@@ -0,0 +1,196 @@
...
+	filters?: ProgramDonorSummaryFilter[];
+};
+
+export type DonorSummaryEntrySort = TableSortOrder & {
+	field: DonorSummaryEntrySortField;
+};
+
+export type DonorSummaryEntrySortField = keyof DonorSummaryRecord;
+
+export type ProgramDonorSummaryEntryField =
```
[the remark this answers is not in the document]
Comment (ciaranschutte) [reply src/app/(post-login)/submission/program/[shortName]/dashboard/components/DonorDataSummary/types.tsx:145]: This is a huge aspect of the project. Originally we added TS on frontend while using GQL with no generation of types. Then we added code generation but never stuck to it, or refactored the code to use the generated types so what remains is redundant, inconsistent and poor typing from our original TS approach/knowledge.

Great improvement to do full code generation of types but the small bit I started makes it very clear it's not a trivial addition.

src/app/hooks/useTimeout.tsx
ciaranschutte: The only change I've made here is move it to a common more descriptive (clear it's a hook) file at a higher level (avoids nasty relative imports)

I'm not clear on what this is doing. This PR is copy/paste from `platform-ui` as we keep saying in our planning.
I'm hoping to clean it up a bit before deployed fully.
At best this appears like it's just making the UI look a little nicer instead of flashing of content.. at worse I think it's an odd workaround for how we're using hooks in places. Requires more work.

justincorrigible (approved):
Approved without testing, copy paste from platform.
May need refactoring later on 👍

PR #167, opened 2024-02-24
Sync http gql

ciaranschutte:
### Sample reg page
- error states are not persisted server side, success states are
- we use a mixture of regular HTTP calls and GQL calls
- on success we can simply refetch and sync with successful state on server
- on error we need to manually update our local apollo cache (state) 
  - cache needs to take same shape as GQL types, server is no longer providing same shape in HTTP calls
  
  
- Functions added for this use case - I think there's more cases where this is happening - keeping this PR small and easy to review

- There is still an issue with error state when navigating back to the sample reg page but this is because of Auth state ( I think ) causing app level re-renders. This PR fixed page level state issues.

### DIFF
--- a/src/app/(post-login)/submission/program/[shortName]/sample-registration/page.tsx
+++ b/src/app/(post-login)/submission/program/[shortName]/sample-registration/page.tsx
@@ -101,27 +101,11 @@ const Register = ({ shortName }: { shortName: string }) => {
 	const registrationId = get(clinicalRegistration, 'id', '') || '';
 
 	/**
-	 * new HTTP endpoints don't return the same data as strictly typed gql endpoints
+	 * new HTTP endpoints don't return the same objects as strictly typed gql endpoints
 	 * data shape needs to match so state can stay in sync on the frontend
 	 */
 	type RequiredProperty = { key: string; accessor: string | (() => void); defaultValue: any };
 	const requiredProperties: RequiredProperty[] = [
-		{ key: 'creator', accessor: 'registration.creator', defaultValue: '' },
-		{ key: 'fileName', accessor: 'fileName', defaultValue: '' },
-		{ key: 'createdAt', accessor: 'registration.createdAt', defaultValue: '' },
-		{ key: 'newDonors', accessor: 'newDonors', defaultValue: [] },
-		{ key: 'newSamples', accessor: 'newSamples', defaultValue: [] },
-		{ key: 'newSpecimens', accessor: 'newSpecimens', defaultValue: [] },
-		{ key: 'alreadyRegistered', accessor: 'alreadyRegistered', defaultValue: [] },
-		{
-			key: 'records',
-			accessor: (resp) =>
-				resp.registration?.records.map((record, index) => ({
-					row: index,
-					fields: Object.keys(record).map((key) => ({ name: key, value: record[key] })),
-				})),
-			defaultValue: [],
-		},
 		{
 			key: 'errors',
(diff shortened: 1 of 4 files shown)


src/app/(post-login)/submission/program/[shortName]/sample-registration/page.tsx
```diff
@@ -100,22 +100,82 @@ const Register = ({ shortName }: { shortName: string }) => {
...
+			onSuccess: async (data, variables, context) => {
+				// succesful uploads will have state on the server
+				if ([200, 201].includes(data.status)) {
+					refetch();
+				} else {
+					// errors are not persisted server side - manually update cache (state store)
+					const registration = await data.json();
+					const registrationUpdate = satisfyGQL({ apiResp: registration, requiredProperties });
+
+					updateClinicalRegistrationQuery((previous) => {
```
Comment (demariadaniel) [inline src/app/(post-login)/submission/program/[shortName]/sample-registration/page.tsx:164]: Will handlers like this be required for other requests?

Comment (ciaranschutte) [reply src/app/(post-login)/submission/program/[shortName]/sample-registration/page.tsx:164]: Yes. I believe so. With that in mind, I wrote these in a way that it should be easy to reuse

Comment (ciaranschutte) [reply src/app/(post-login)/submission/program/[shortName]/sample-registration/page.tsx:164]: Anything that is:
- not a GQL response
- doesn't have state server side (ie. can't invalidate and get new data with Apollo `refetch`)

Comment (demariadaniel) [reply src/app/(post-login)/submission/program/[shortName]/sample-registration/page.tsx:164]: JS Docs are good, wondering if it's worth documenting in the readme, so there's a guide for future features where this is required

src/app/(post-login)/submission/program/[shortName]/sample-registration/page.tsx
```diff
@@ -100,22 +100,82 @@ const Register = ({ shortName }: { shortName: string }) => {
 	const hasErrors = !!schemaOrValidationErrors.length;
 	const registrationId = get(clinicalRegistration, 'id', '') || '';
 
+	/**
+	 * new HTTP endpoints don't return the same objects as strictly typed gql endpoints
+	 * data shape needs to match so state can stay in sync on the frontend
+	 */
```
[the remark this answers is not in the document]
Comment (ciaranschutte) [reply src/app/(post-login)/submission/program/[shortName]/sample-registration/page.tsx:106]: Strictly this part, where we massage the API response to match the GQL shape, could be done on the backend.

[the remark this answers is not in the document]
Comment (demariadaniel) [reply src/app/(post-login)/submission/program/[shortName]/sample-registration/page.tsx:106]: Seems like a lot of maintenance for the front end, especially considering we have functions on the Clinical side specifically intended to massage the response:

https://github.com/icgc-argo/argo-clinical/blob/develop/src/schemas/utils.ts#L258

https://github.com/icgc-argo/argo-clinical/blob/develop/src/schemas/clinical-mutations/uploadClinicalSubmissions.ts

So it might be worth reviewing what needs to change on the back end creating a ticket to address, at least to minimize future work

demariadaniel (approved):

src/app/(post-login)/submission/program/[shortName]/sample-registration/page.tsx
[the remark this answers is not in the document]
Comment (ciaranschutte) [reply src/app/(post-login)/submission/program/[shortName]/sample-registration/page.tsx:106]: I agree.

The new endpoints should not have changed the data shape or it should have been communicated & frontend state implications discussed. 

Endpoint responses could be rewritten. That would be ideal.
But the frontend is blocked until that's done without this.

PR #173, opened 2024-04-14
Cleanup clinical entity data table

ciaranschutte:
**Context**: 
Lots of code "copy/pasted" from `platform-ui` have TS types written be developers.
These clash with generated types from GQL endpoints.
The GQL generated types is the correct way, as our written don't always encompass the types correctly and are prone to error.

Additionally we have many places that have empty objects as default values. These cause type errors when they differ from the actual returned object. Example, a filter function is expecting `ClinicalData[]` but it actually gets `EmptyClinicalData[]` that doesn't quite satisfy type constraints. This leads to writing code to type narrow/account for both of these types OR making them satisfy the same type. Array methods in particular are tricky.

Another difference is there is a major TS version change between `platform-ui` and `rdpc-ui` so the "copy/paste" approach from `platform-ui` surfaced a lot of errors. ( v4 vs v5 )

**Changes:**
This PR mostly changes types, fixes a couple of generics and adds extra property to satisfy types.
A couple of instances of using `any` because it's a bigger issue that needs to be fixed at a component level.
For clinical data page a lot of code reorganisation to solve type issues by type narrowing in correct place eg. if no data/loading, there's often a lot of code we don't ever need to worry about unless we _do_ have data or are _not_ loading.

Disables ESLint on build. We can re-enable later when repo is in better shape. For now step 1 is having project successfully build.


ciaranschutte:
@joneubank tagging for visibility as I know you were looking at this also. let's avoid redundant work.

src/app/(post-login)/submission/program/[shortName]/clinical-data/common.tsx
```diff
@@ -209,8 +211,9 @@ export const hasClinicalErrors = (
 			),
 	).length > 0;
 
-export const emptyClinicalDataResponse: ClinicalEntityQueryResponse = {
+export const emptyClinicalDataResponse: SubmittedDataSideMenuQuery = {
 	clinicalData: {
+		programShortName: '',
```
Comment (ciaranschutte) [inline src/app/(post-login)/submission/program/[shortName]/clinical-data/common.tsx:216]: not ideal but works ok as a type-fix - this is just an empty object meant as a default value to display nothing - "emtptyClinicalDataResponse"

the root cause is the mismatch between TS types and GQL generated types.

joneubank (changes requested):
As discussed offline, just need to remove the `any` usage and this is fine.

I left one other question, let me know your thoughts... no change needed.

src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable/index.tsx
```diff
@@ -0,0 +1,636 @@
...
+							tableProps={{
+								page: errorPage,
+								pages: numErrorPages,
+								pageSize: errorPageSize,
+								sorted: errorSorted,
+								onPageChange: (value) => updatePageSettings('page', value),
+								onPageSizeChange: (value) => updatePageSettings('pageSize', value),
+								onSortedChange: (value) => updatePageSettings('sorted', value),
+								// TODO: Test + Update Pagination in #2267
+								// https://github.com/icgc-argo/platform-ui/issues/2267
```
Comment (demariadaniel) [inline src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable/index.tsx:601]: This Todo can be removed

Comment (ciaranschutte) [reply src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable/index.tsx:601]: thank you - the exact reason I'm against putting TODOs in code

src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable/index.tsx
```diff
@@ -0,0 +1,636 @@
...
+		pageSize,
+		sort,
+		completionState,
+		donorIds,
+		submitterDonorIds,
+	);
+
+	const { clinicalData } = clinicalEntityData;
+
+	// TODO: can you have ClinicalEntityErrors without ClinicalEntityData?
```
Comment (demariadaniel) [inline src/app/(post-login)/submission/program/[shortName]/clinical-data/ClinicalEntityDataTable/index.tsx:196]: Technically yes ClinicalEntityErrors is supposed to work as a separate standalone query, so this TODO can be removed.
However I tested and it is not working in Develop. I have a patch PR coming in shortly.

ciaranschutte:
discussed offline, adding back in ErrorNotification type. build will remain broke but will address full fix in seperate PR

joneubank (approved):

PR #177, opened 2024-04-21
App reorg

ciaranschutte:
- `app` folder just for page level routing + special files eg. `layout` 
  - fully supported ref: https://nextjs.org/docs/app/building-your-application/routing/colocation#store-project-files-outside-of-app
- moves components to components folder
- moves "instances" of components + data to `views` similar to OHCRN


src/hooks/AppProvider.tsx
demariadaniel, line 1: Manually reviewing files, `import { AppConfig }` should be updated to `import type { AppConfig } from '@/app/api/config/config';` since we are declaring `const AppConfig` @ L42

ciaranschutte: Sounds good - nice catch - this isn't a change in this PR. It'll say "File renamed without changes" on this file with no visual diff. Confirmed it's in `develop` already but doesn't throw a type error.

demariadaniel (approved):

PR #178, opened 2024-04-21
Next14 upgrade

ciaranschutte:


package.json
```diff
@@ -28,7 +28,7 @@
     "graphql": "^16.8.0",
     "js-cookie": "^3.0.5",
     "lodash": "^4.17.21",
-    "next": "^13.4.19",
```
justincorrigible, line 31: based on it being the only change in this PR, I take it this update is seamless?
nothing breaking sees surprising for a major version update.

ciaranschutte: Yessir. checked the breaking changes in the changelog and verified locally

justincorrigible (approved):

PR #184, opened 2024-05-07
Nav bugs

ciaranschutte:
- couple of errors in components with data fetching states causing pages to crash
- gql codegen pointing at old gateway with types, regenerating codegen types

### DIFF
--- a/codegen.ts
+++ b/codegen.ts
@@ -27,7 +27,7 @@ const gqlConfig: CodegenConfig = {
 		'./src/__generated__/gateway/': {
 			documents: ['src/**/gql/gateway/*.ts'],
 			preset: 'client',
-			schema: `${process.env.NEXT_PUBLIC_GATEWAY_API_ROOT}/graphql`,
+			schema: `${process.env.NEXT_PUBLIC_GATEWAY_API_ROOT}`,
 			presetConfig: {
 				gqlTagName: 'gql',
 			},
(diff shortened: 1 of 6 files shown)


src/views/submission/dashboard/components/ClinicalChart/index.tsx
```diff
@@ -70,7 +70,6 @@ const useProgramDonorPublishedAnalysisByDateRangeQuery = (
 	const pollingTimeout = useTimeout(30000);
 	const hook = useGatewayQuery(PROGRAM_DONOR_PUBLISHED_ANALYSIS_BY_DATE_RANGE_QUERY, {
 		...options,
-		skip: !dateRangeFrom || !dateRangeTo,
```
ciaranschutte, line 73: unclear why we're skipping the query here - no comment - state from above only updates selectively from a useEffect dependency

codegen.ts
```diff
@@ -27,15 +27,15 @@ const gqlConfig: CodegenConfig = {
...
+			schema: `${process.env.NEXT_PUBLIC_GATEWAY_API_ROOT}`,
 			presetConfig: {
 				gqlTagName: 'gql',
 			},
 		},
 		'./src/__generated__/clinical/': {
 			documents: ['src/**/gql/clinical/*.ts'],
 			preset: 'client',
-			schema: `${process.env.NEXT_PUBLIC_CLINICAL_API_ROOT}/graphql`,
+			schema: `${process.env.NEXT_PUBLIC_CLINICAL_API_ROOT}`,
```
ciaranschutte, line 38: new rdpc work uses `/gql` instead of platform `/graphql`
remove hardcoding and use full url from env var

justincorrigible (approved):
LGTM

PR #185, opened 2024-05-07
Endpoint fixes

ciaranschutte:
- make endpoint and gql endpoint into distinct env vars
- hardcoding the `gql` or `graphql` path in code requires the backends to stick to consistent naming, which is currently broke
- allows UI to work if convention on backend isn't followed


src/app/api/config/config.ts
```diff
@@ -80,11 +82,19 @@ export const getAppConfig = (serverEnv: any): AppConfig => {
...
+			'https://argo-gateway.argo-dev.cumulus.genomeinformatics.org/graphql',
 		CLINICAL_API_ROOT:
 			serverEnv.NEXT_PUBLIC_CLINICAL_API_ROOT ||
 			process.env.NEXT_PUBLIC_CLINICAL_API_ROOT ||
-			'https://argo-gateway.dev.argo.cancercollaboratory.org',
+			'https://clinical.argo-qa.cumulus.genomeinformatics.org',
+		CLINICAL_API_GRAPHQL:
+			serverEnv.NEXT_PUBLIC_CLINICAL_API_GRAPHQL ||
+			process.env.NEXT_PUBLIC_CLINICAL_API_GRAPHQL ||
+			'https://clinical.argo-qa.cumulus.genomeinformatics.org/gql',
```
joneubank, line 97: Please don't commit defaults that connect to specific environments. I would prefer a default that intentionally fails to connect. We don't want a misconfigured prod accidentally connecting to a lower env. We also don't need to be broadcasting our internal URLs.

justincorrigible, line 97:
```suggestion
			'http://local',
```

justincorrigible (approved):

